PR #14, opened 2026-08-17
refactor(backend): close the gap between the described and the enforced architecture

NaimElijah:
## Problem

ADR-001 states that architecture violations fail the build, and `backend/CLAUDE.md` tells contributors
to "treat that test as the definition of the module boundaries rather than working from a layout
described in prose."

A conformance audit found the structure sound in its bones — a genuinely framework-free domain,
hand-wired pure domain services, correct port/adapter direction on the driven side, package-private
Spring Data repositories, `@Transactional` confined to application services — but the enforcement
weaker than the documents claimed. Five ArchUnit rules covered roughly half of ADR-001's commitments,
and the bounded-context rule's ignore list matched **every** cross-context edge that existed, so it
could not fail. Contributors were being told to trust a guard that was partly decorative.

Behind that gap sat one dependency cycle, an inverted application→adapter dependency in six of eight
contexts, a shared kernel holding other contexts' events, an aggregate whose invariants could be walked
around, and five behavioural bugs.

## Approach

Fix the code first, then tighten each rule only once the code already satisfied it, then correct the
documents to match. Commits are sequenced so each stands alone and each bug fix begins with a failing
test.

## What was done

**Bugs** (each red before green, one commit apiece)

| Fix | Was |
|---|---|
| `f4bd1c3` | `changeDifficulty` promoted an ACTIVE upgrade to HARD, bypassing the max-3-concurrent rule |
| `1ff7e53` | NUMERIC progress compared bare numbers, so 45 minutes cleared a 30 km target |
| `a3b7701` | `UpgradeOverdueDetected` had a listener but was never published |
| `8d63479` | `NotificationScheduler` injected a `Clock`, then called the overload that read `LocalDate.now()` |
| `e5e4951` | Reviving an ABANDONED upgrade to PLANNED announced nothing |

**Boundaries**

- **Cycle broken** (`bb78ad0`). `upgrade ⇄ tracking` was formed entirely from individually-sanctioned
  edges. `upgrade` now declares `UpgradeTrackingSummaryPort` — what it needs, in a record it owns — and
  `tracking` supplies it, so both arrows run the same way.
- **Application layer turned away from the web** (`8a92bd9`). Services take command records from
  `application/port/in` and return domain objects; a `*WebMapper` per context translates at the
  boundary. This makes the pattern `upgrade` and `dashboard` already used the rule, not the exception.
- **Events moved to the contexts that raise them** (`ca95179`). Only the `DomainEvent` marker stays
  shared; `DomainEventPublisher` moves to `common/domain/port/out` where the outbound-ports rule finally
  covers it.
- **Aggregates closed** (`dceb92e`, `63c2a92`). `HealthUpgrade` and `Reminder` lose their setters;
  `HealthUpgrade` gains a validating `create`, and `Reminder` gains `ReminderDays` + `isDueAt`, taking
  behaviour back from an adapter in another context that was re-parsing its storage format.

**Enforcement** (`7eb2d2b`) — five rules become ten: cycle freedom, application must not depend on
**any** adapter, inbound adapters must not depend on outbound ones, inbound ports must be interfaces,
controllers only in the web adapter, Spring Data repositories package-private. Domain purity widens
beyond Spring to bean validation, servlet, Jackson and Hibernate.

**Safety net** (`c319aca`) — `ApplicationContextIT` boots the app against real PostgreSQL, catching a
lost `@Bean` in the hand-wired `*BeansConfig` classes and a missing Flyway migration. Tests added for
`UpgradeSchedulingService` and `GlobalExceptionHandler`. JaCoCo reports coverage; OWASP
dependency-check and `npm audit` fail on known vulnerabilities.

**Documents** (`60f855b`, `3d8fdbe`) — ADR-002 records the audit and the corrections; ADR-001 is left
intact as history with its status marked superseded in part.

## Trade-offs accepted

- More types: a command/result record and a mapper per context. Deliberate — it is what makes the
  dependency direction enforceable rather than aspirational.
- `UpgradeTrackingConfigDto` duplicates the field list of `TrackingConfigDto`. That is the price of each
  context owning its own contract; a rename on one side no longer breaks the other.
- Enum values cross the upgrade/tracking seam as their names, losing compile-time typing at that one
  point. The JSON is unchanged, because that is what Jackson already emitted.
- `mvn verify` now needs a database. `mvn test` deliberately still does not.
- The JPA-annotated-domain compromise from ADR-001 **stays**; this PR verifies the domain is otherwise
  honest rather than splitting persistence models across nine contexts.

## How it was verified

- `mvn test` green after every phase: **63 tests at baseline → 129**.
- `mvn verify` green against a real PostgreSQL: 4 Flyway migrations applied, `ddl-auto: validate`
  accepted the schema, hand-wired beans resolved.
- **Every new ArchUnit rule was verified to bite** — the violation each targets was reintroduced, the run
  confirmed red for that specific rule, and the change reverted. A rule that has never failed is not
  known to work.
- The upgrade response JSON contract is pinned by a test (field names, order, enum rendering, null
  handling) and matches `frontend/src/types/index.ts`.
- Frontend untouched: `git diff main...HEAD -- frontend/` is empty.

## Known gaps

- The full stack was never run end to end (`docker-compose up --build`); verification is the integration
  test plus the serialisation contract test.
- The new CI audit steps have never executed. OWASP dependency-check needs `NVD_API_KEY` in repo
  secrets and may fail or run slowly on first use.
- No Surefire guard against the architecture suite silently ceasing to run: `failIfNoTests` only applies
  with `-Dtest`, so it would be reassurance rather than protection. Recorded in ADR-002.

## ADRs

- [ADR-002 — Close the gap between the described and the enforced architecture](docs/ADRs/ADR-002-close-the-gap-between-the-described-and-enforced-architecture.md)
- [ADR-001 — DDD + Hexagonal](docs/ADRs/ADR-001-ddd-hexagonal-architecture.md) (superseded in part)


NaimElijah:
# Code review

Scope: `git diff origin/dev...HEAD` — 80 files, +2500/−578.

## Verified clean

- The headline HARD-limit fix is correct on every route traced (activate from PLANNED/PAUSED, promote
  ACTIVE, promote non-ACTIVE, no-op update). `countByUserIdAndStatusAndDifficulty` never double-counts
  the subject upgrade, because `difficultyChanges` is false when it is already HARD.
- No `application` class imports an adapter anywhere, and `common` imports no bounded context — both new
  structural rules genuinely hold.
- The `UpgradeDto` JSON contract is unchanged: enum→String preserves names and field order, absent
  config still renders `null`.
- Every new `Clock` injection resolves against the existing bean.

## Findings

### 1. `ReminderDays` — an all-invalid day list becomes "every day" (high)

`reminder/domain/model/ReminderDays.java:53`

`of(["Mondays","Fridays"])` passes the null/empty short-circuit, both tokens fail `parse`, the set ends
up empty, and `includes()` treats empty as *every day*. A typo'd filter fires the reminder daily instead
of twice weekly. The old code stored tokens verbatim and never matched, so this **flips the failure from
too-few to too-many**. `days_ignoreBlanksAndUnrecognisedTokens` misses it because it includes one valid
token.

### 2. Event payloads with user content are logged at DEBUG (high)

`common/adapter/out/event/SpringDomainEventPublisher.java:29`

The new central log line emits whole records via `toString()`. `HealthUpgradeCreated` carries the
user-authored title next to a userId, and `application.yml` sets `com.healthupgrades: DEBUG`. In a health
app those titles are personal data. The comment presents the generic form as an advantage, which means
every future event field is logged automatically with no review. Against CLAUDE.md: *"Never log
credentials, tokens, or personal data."*

### 3. `HealthUpgrade` lost its status default without closing the builder (medium)

`upgrade/domain/model/HealthUpgrade.java:127`

`@PrePersist` no longer defaults `status` to IDEA, but the `@Builder` is still public and documented as
the JPA/test construction path. A `builder()...build()` that omits status now reaches a `nullable = false`
column and fails at flush, surfacing as a 500 far from the call site. Previously it was silently repaired.
Make it structural — `@Builder.Default` — or close the builder.

### 4. `onOverdue` pays for a lookup it then discards, daily and forever (medium)

`notification/adapter/in/event/NotificationEventListener.java:101`

`title(...)` is an argument, so it is evaluated before `createOncePerUpgrade` can short-circuit. A
permanently-overdue upgrade is rediscovered by the scan every day, costing one wasted `findOwned` per
upgrade per day indefinitely. The removed `notifyOverdue` had the aggregate in hand and checked `exists`
first.

### 5. `application/port/out` is covered by no architecture rule (medium)

`upgrade/application/port/out/UpgradeTrackingSummaryPort.java:29`

`outbound_ports_are_interfaces` matches only `..domain.port.out..`; `inbound_ports_are_interfaces` only
`..application.port.in..`. The port the entire cycle-break rests on is unguarded, in a PR whose thesis is
that documented boundaries must be machine-checked.

### 6. `Notification` gained `markAsRead()` but kept `@Setter` (medium)

`notification/domain/model/Notification.java:49`

`setRead(boolean)` is still public, so the encapsulation the method advertises is not enforced — unlike
`HealthUpgrade` and `Reminder`, which actually lost their setters. The aggregate is no more closed than
before, only larger.

### 7. `NOTIFY_OVERDUE_CRON` renamed with no fallback (medium)

`backend/src/main/resources/application.yml:32`

Silently renamed to `UPGRADE_OVERDUE_CRON`. A deployment setting the old name loses its override and
reverts to the 08:00 default with no warning — while ADR-002 asserts "no behaviour change beyond the five
bug fixes". A chained default keeps both working.

### 8. `TrackingConfigQuery.findByUpgradeId` is now dead published surface (medium)

`upgrade/adapter/in/web/UpgradeWebMapper.java:40`

`UpgradeWebMapper` was its only consumer and now goes through `UpgradeTrackingSummaryPort`. A port method
nothing exercises stays in tracking's published contract and will be maintained as if it were live.

### 9. A web adapter drives an outbound port with no use case involved (medium)

`upgrade/adapter/in/web/UpgradeWebMapper.java:28`

Cross-context fetching now happens in `adapter.in.web`. A second driving adapter needing upgrade+tracking
must duplicate the composition or depend on a web mapper. `inbound_adapters_do_not_depend_on_outbound_adapters`
bans only `adapter.in → adapter.out`, so nothing catches this.

### 10. `npm audit --audit-level=high` blocks the merge path on dev-tool advisories (medium)

`.github/workflows/ci.yml:76`

It audits dev dependencies, so an advisory anywhere in the Vite/ESLint toolchain reddens every unrelated
PR with no bearing on shipped code. `--omit=dev` scopes it to what ships.

### 11. Clock-independence proven with the system clock (medium)

`tracking/domain/service/StreakCalculatorTest.java:92`, `upgrade/adapter/in/scheduling/UpgradeOverdueSchedulerTest.java:107`

Both assertions that prove the code ignores the wall clock are themselves written against
`LocalDate.now()`. Run the suite on the wrong date and they fail. Against CLAUDE.md: *"Deterministic
always."* A second fixed date proves the same point.

### 12. `healtharea` has two names for one query (low)

`healtharea/application/HealthAreaService.java:36`

`findAll(userId)` and `listByUser(userId)` are now byte-identical; the DTO mapping that distinguished them
moved to the web mapper. A future change to scoping or ordering must be made twice or they diverge.

### 13. `ReminderDays` polish (low)

`reminder/domain/model/ReminderDays.java:63`

Hand-rolled `reduce` where `Collectors.joining` exists; no `equals`/`hashCode` on a value type, while
`Reminder.days()` mints a fresh instance per call; `DayOfWeek.values()` cloned per token and the CSV
re-parsed on every `isDueAt`, in a job that runs every minute over every enabled reminder.

## Discussed and not changed

**The unit-comparability guard rarely fires.** Raised as "the 45-minutes-vs-30-km defect is not closed",
because the UI never sends `entry.unit` and a blank `targetUnit` also short-circuits. The semantics are
deliberate: an entry that states no unit *is* in the configured unit — that is what the numeric input,
labelled with the target unit, means. A stricter rule would reject every entry the current UI produces.
The guard exists to catch the case where a unit is stated and disagrees. Making units mandatory is a
product decision, not a correctness fix.

PR #21, opened 2026-08-18
fix: align the frontend/backend enum contract and the dashboard completion rate

### LINKED ISSUE #19: fix: five of the eight upgrade types the UI offers are rejected by the API with a 500
## Verified

Confirmed on `dev` at `ca3391e`, by reading both sides and by probing the API with a standalone
MockMvc request.

**Backend** (`upgrade/domain/model/UpgradeType.java`) accepts exactly four values:

```
HABIT, GOAL, EXPERIMENT, PROTOCOL
```

**Frontend** (`src/types/index.ts`) declares eight, and every one of them is offered in the
create-upgrade form (`UpgradeBacklogPage.tsx`, `typeOptions`):

```
HABIT, ONE_TIME_ACTION, PRODUCT_REPLACEMENT, ROUTINE,
GOAL, EXPERIMENT, LEARNING_TASK, MEDICAL_PREVENTIVE
```

So five of the eight choices a user can pick are values the API cannot bind. `PROTOCOL` runs the
other way: the backend accepts it, nothing in the frontend or the README mentions it.

The same drift exists for `Frequency`: the backend has `DAILY, WEEKLY, MONTHLY`; the frontend type
adds `CUSTOM`, and `UpgradeDetailsPage` offers it in the tracking-config form. There is no custom
schedule concept behind it — nothing in the backend would know what to do with one.

## Impact

Picking any of the five unsupported types returns **HTTP 500**, not 400:

```
POST /api/upgrades  {"title":"x","type":"ONE_TIME_ACTION"}
→ 500 {"status":500,"message":"Internal server error",...}
```

Jackson cannot bind the unknown enum constant and throws `HttpMessageNotReadableException`, which is
caught by the catch-all `@ExceptionHandler(Exception.class)` in `GlobalExceptionHandler` before
Spring's default 400 handling can apply. The user sees a generic failure with nothing to act on, and
the server logs an unhandled error for what is really invalid input.

Five of the eight upgrade kinds the README advertises are therefore unusable.

## Which side is wrong

The README defines the product as eight kinds of upgrade — "a new habit, a one-time action, a product
replacement, a routine, a goal, an experiment, a learning task, a medical/preventive task" — which is
exactly the frontend's list. The backend enum is the outlier, and `PROTOCOL` appears in no
requirement, no seed row and no UI.

So: widen the backend to match the documented product, rather than narrowing the frontend to match an
enum nobody wrote down.

`Frequency` is the opposite case — `CUSTOM` is backed by nothing on either side — so there the
frontend is the outlier and loses the value.

## Proposed fix

- Add `ONE_TIME_ACTION`, `PRODUCT_REPLACEMENT`, `ROUTINE`, `LEARNING_TASK`, `MEDICAL_PREVENTIVE` to
  the backend `UpgradeType`.
- Keep `PROTOCOL` — a deployed database may hold rows with it, and Hibernate would fail to read them
  back if the constant disappeared. Add it to the frontend type and badge map so such a row renders,
  but leave it out of the create form, since it is not one of the documented kinds.
- Remove `CUSTOM` from the frontend `Frequency`.
- Pin the agreement with a test, so the next divergence fails the build instead of reaching a user.

No migration is needed: `health_upgrades.type` and `tracking_configs.frequency` are `VARCHAR(50)` with
no CHECK constraint, so widening the enum changes no schema.

## Acceptance criteria

- [ ] Every type offered in the create form is accepted by the API
- [ ] `Frequency` matches on both sides
- [ ] A test fails if the backend enum and the frontend union drift apart again
- [ ] `mvn test` and `npm run build` pass

### LINKED ISSUE #20: fix: dashboard weekly completion rate is rendered 100x too large
## Verified

Confirmed on `dev` at `ca3391e` by reading both sides of the contract.

**The backend already returns a percentage.** `DashboardAggregationService.calculateWeeklyCompletionRate`:

```java
return (double) completed / weekEntries.size() * 100.0;
```

and `DashboardAggregationServiceTest` pins it — for three entries of six completed it asserts
`50.0`, not `0.5`.

**The frontend scales it again.** `DashboardPage.tsx`:

```ts
const completionPct = Math.round((data?.weeklyCompletionRate ?? 0) * 100);
```

## Impact

A week in which half the logged entries were successful is displayed as **5000%**, not 50%.

The value is rendered in three places on the dashboard, all wrong by the same factor:

- the "Weekly Rate" stat tile
- the progress bar's width, as `style={{ width: '5000%' }}` — the bar is always pinned full for any
  non-zero rate, so it carries no information at all
- the numeric label beside the bar

Only a genuinely zero rate displays correctly, which is why it survived: an account with no progress
logged shows `0%` and looks fine.

`weeklyCompletionRate` has exactly one consumer, so nothing else depends on the wrong reading.

## Proposed fix

Drop the extra `* 100` in `DashboardPage.tsx` and round the value the API already sends. Record the
unit on the shared type so the next reader does not have to guess whether it is a fraction or a
percentage.

The backend needs no change — its side of the contract is already correct and already tested.

## Note on testing

The standard is that a bug fix starts with a failing test. The frontend has **no test runner** — no
vitest, no jest, nothing in `devDependencies` — so there is no way to write one without first
introducing a test framework, which is a larger decision than this fix and deserves its own ADR.

The backend half of the contract is pinned by the existing `DashboardAggregationServiceTest`. The
frontend half is verified by inspection and by the type's documented unit.

## Acceptance criteria

- [ ] A 50% week renders as `50%`
- [ ] The progress bar width tracks the real rate
- [ ] The unit is documented on `DashboardDto.weeklyCompletionRate` in `src/types`
- [ ] `npm run build` and eslint pass

NaimElijah:
Closes #19
Closes #20

## Problem

Two contract mismatches between the frontend and the backend, both found while documenting the
project in #18 and both verified before being acted on.

**1. Five of the eight upgrade types the UI offers were rejected (#19).** The backend enum had
`HABIT, GOAL, EXPERIMENT, PROTOCOL`; the frontend offered eight, five of which the API could not
bind. Verified with a standalone MockMvc request:

```
POST /api/upgrades  {"title":"x","type":"ONE_TIME_ACTION"}
→ 500 {"status":500,"message":"Internal server error",...}
```

500, not 400 — Jackson's `HttpMessageNotReadableException` is caught by the catch-all
`@ExceptionHandler(Exception.class)` before Spring's default 400 handling applies. `Frequency` had
the same shape of problem with a `CUSTOM` the backend never had.

**2. The weekly completion rate was rendered 100× too large (#20).** The backend returns a
percentage — `completed / total * 100.0`, pinned at `50.0` by `DashboardAggregationServiceTest` — and
`DashboardPage` multiplied by 100 again. A 50% week displayed as `5000%`, and the progress bar was
rendered with `width: 5000%`, so it sat full for any non-zero rate.

## Approach

**For the enum drift, the backend was the side that was wrong.** The README defines the product as
exactly the eight kinds the frontend offers; `PROTOCOL` appears in no requirement, no seed row and no
screen. So the enum was widened to match the documented product rather than five features being
deleted from the UI.

`PROTOCOL` is kept anyway: a deployed database may hold rows carrying it, and Hibernate would fail to
read those back if the constant disappeared. It renders on such a row but is deliberately absent from
the create form.

`CUSTOM` is the opposite case — nothing on either side implements a custom schedule, and a per-day
schedule is what reminders are for — so the frontend loses it.

**The agreement is now pinned.** `FrontendEnumContractTest` reads `frontend/src/types/index.ts` and
asserts each mirrored union matches its backend enum, across all seven mirrored pairs. It fails on
both of these divergences before the fix and passes after.

**For the rate, only the frontend changed** — the backend's half of the contract was already correct
and already tested. The unit is now stated on the shared type.

## Trade-off accepted

- **The rate fix ships without a test of its own.** The frontend has no test runner — no vitest, no
  jest, nothing in `devDependencies` — so a failing-test-first fix would mean introducing a test
  framework, a new dependency and a CI step, which is a bigger decision than a one-line arithmetic
  fix and deserves its own ADR. The backend half is pinned by the existing test; this half is verified
  by inspection and by the documented unit.
- **`FrontendEnumContractTest` reads across module boundaries**, so it is coupled to the frontend
  file's path. That coupling is deliberate — any indirection would be one more thing to keep in step,
  which is the class of problem being fixed.
- **`PROTOCOL` is retained rather than removed**, accepting a value in the enum that no requirement
  asks for, because removing it would break reads on any database already holding it.

## Not fixed here

An unbindable request body returns **500 rather than 400** — the catch-all handler in
`GlobalExceptionHandler` intercepts `HttpMessageNotReadableException` before Spring's default handling
applies. Aligning the enums removes the way users were hitting it, but malformed JSON still produces
a 500. That is a separate defect and a wire-contract change, so it is not bundled in here.

## How it was verified

- `mvn test` — passes, including the new `FrontendEnumContractTest` (8 tests). Confirmed failing on
  both divergences before the fix.
- `npm run build` — passes (`tsc` + Vite).
- `npx eslint src --ext .ts,.tsx` — clean.
- No migration needed: `health_upgrades.type` and `tracking_configs.frequency` are `VARCHAR(50)` with
  no CHECK constraint.

## Related

`docs/architecture/architecture.md` said the hand-written frontend types had nothing detecting drift.
That is now half true and the entry says so: the enums are checked, the DTO field shapes still are
not. No new ADR — this corrects an implementation to match a documented product, it does not decide
anything.

### DIFF
--- a/backend/src/main/java/com/healthupgrades/upgrade/domain/model/UpgradeType.java
+++ b/backend/src/main/java/com/healthupgrades/upgrade/domain/model/UpgradeType.java
@@ -5,11 +5,25 @@ package com.healthupgrades.upgrade.domain.model;
  *
  * <p>The type is descriptive — it drives presentation and filtering, not domain rules. How progress is
  * measured is decided separately by the tracking configuration.
+ *
+ * <p>The first eight values are the kinds the README defines as the product. They and the frontend's
+ * {@code UpgradeType} union must stay identical: the value is bound straight from the request body, so
+ * a name the frontend offers and this enum lacks fails to deserialize. {@code FrontendEnumContractTest}
+ * fails the build when the two drift apart.
  */
 public enum UpgradeType {
 
     /** A recurring behaviour to establish, e.g. drinking two litres of water a day. */
     HABIT,
+
+    /** A single task to get done, e.g. booking a dentist appointment. */
+    ONE_TIME_ACTION,
+
+    /** Swapping something out for a better alternative, e.g. replacing toxic cleaning products. */
+    PRODUCT_REPLACEMENT,
+
+    /** A defined sequence to follow regularly, e.g. a morning meditation and stretch. */
+    ROUTINE,
 
     /** An outcome to reach by a date, e.g. running 10km by summer. */
     GOAL,
@@ -17,6 +31,18 @@ public enum UpgradeType {
     /** A time-boxed trial to learn from, e.g. no caffeine after 2pm for two weeks. */
     EXPERIMENT,
 
-    /** A defined routine or one-off procedure to follow, e.g. a morning stretch sequence. */
+    /** Something to read up on or study, e.g. reading about intermittent fasting. */
+    LEARNING_TASK,
+
+    /** A preventive or medical errand, e.g. getting bloodwork done. */
+    MEDICAL_PREVENTIVE,
+
+    /**
+     * A prescribed protocol to follow.
+     *
+     * <p>Retained rather than offered: it predates the kinds above, appears in no requirement and is not
+     * selectable in the UI, but a deployed database may hold rows carrying it — and Hibernate would fail
+     * to read those back if the constant were removed.
+     */
     PROTOCOL
 }
--- a/backend/src/main/java/com/healthupgrades/upgrade/domain/model/UpgradeType.java
+++ b/backend/src/main/java/com/healthupgrades/upgrade/domain/model/UpgradeType.java
@@ -43,6 +43,10 @@ public enum UpgradeType {
      * <p>Retained rather than offered: it predates the kinds above, appears in no requirement and is not
      * selectable in the UI, but a deployed database may hold rows carrying it — and Hibernate would fail
      * to read those back if the constant were removed.
+     *
+     * @deprecated not one of the documented upgrade kinds. Existing rows keep it and still render; new
+     *             code should choose from the eight above. Remove once no stored row carries it.
      */
+    @Deprecated
     PROTOCOL
 }
--- a/backend/src/test/java/com/healthupgrades/contract/FrontendEnumContractTest.java
+++ b/backend/src/test/java/com/healthupgrades/contract/FrontendEnumContractTest.java
@@ -52,6 +52,13 @@ class FrontendEnumContractTest {
     private static final Pattern UNION_MEMBER = Pattern.compile("'([A-Z_]+)'");
 
     /**
+     * Matches a union of screaming-caps string literals, which is how this codebase spells a mirrored
+     * backend enum and nothing else.
+     */
+    private static final Pattern MIRRORED_UNION = Pattern.compile(
+            "export type (\\w+)\\s*=\\s*((?:\\s*\\|?\\s*'[A-Z_]+')+)\\s*;", Pattern.DOTALL);
+
+    /**
      * Every backend enum that the frontend mirrors, paired with the union that mirrors it.
      *
      * <p>A new mirrored enum belongs here; one that is not listed is not checked.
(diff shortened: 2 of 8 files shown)


NaimElijah (commented):
## Review

Self-review of my own PR, so treat it as a checklist rather than a second opinion. Two things I think
should change before merge, and four risks I checked and found clear.

---

### 1. The guard is itself hand-maintained — `FrontendEnumContractTest`

`mirroredEnums()` is a hard-coded list of the seven pairs. Add a new enum to the backend, mirror it in
`src/types/index.ts`, forget to register it here, and it is simply not checked — which is the same
failure mode this test exists to prevent, moved one level up. `GivenTheFrontendTypesFile_...` catches a
**rename** of something already listed, but nothing catches an **unlisted** new pair.

Worth closing, because the cost of getting it wrong is exactly the bug being fixed. Deriving the
expected set from the frontend file — every union whose members are screaming-caps is by definition a
mirrored enum — and failing when one is unregistered makes the guard self-maintaining.

### 2. `PROTOCOL` is "retained, not offered" by comment only

The doc comment says it is legacy and must not be used for new upgrades, and nothing enforces that.
Marking the constant `@Deprecated` turns the convention into a compiler warning at every new use site,
which is where it belongs.

Checked that this is safe: no `switch` over `UpgradeType` exists anywhere in the backend, so no
exhaustiveness warning appears, and `values()` in the contract test does not warn.

---

### Risks checked, no change needed

**Reordering the enum cannot corrupt stored data.** Every `@Enumerated` in the backend is `STRING`;
there is no `ORDINAL` mapping, so the constants' order is not persisted and inserting five values in
the middle changes nothing already written.

**The cross-module file read does not break the image build.** `FrontendEnumContractTest` reads
`../frontend/src/types/index.ts`, and `backend/Dockerfile` copies only `pom.xml` and `src` — but it
runs `mvn package -DskipTests`, so the test never executes there. CI checks out the whole repository,
so the path resolves. The upward search from the working directory covers being run from either the
repository root or `backend/`.

**Removing `CUSTOM` cannot orphan existing data.** For a row to hold it, the API would have had to
accept it, and it never could — the backend enum has never had the constant, so binding always failed.
No migration, no backfill.

**No other UI surface lists these values.** `typeOptions` in `UpgradeBacklogPage` is the only place
upgrade types are enumerated for input, and the frequency select in `UpgradeDetailsPage` is the only
one for frequency. There is no type filter elsewhere that would still offer a rejected value.

### Nit, not worth a change

`PROTOCOL` and `EXPERIMENT` both render as a grey badge, so the two are visually identical. Acceptable
while `PROTOCOL` is legacy and effectively absent from real data; if it ever becomes common it needs
its own colour.

---

### On the scope boundary

Leaving the 500-on-unbindable-body defect out of this PR is right — it is a wire-contract change with
its own blast radius — but it does mean the underlying sharp edge survives this fix. Anything that
sends malformed JSON still gets a 500 instead of a 400. That deserves its own issue rather than being
left in a PR description.

NaimElijah:
Review findings addressed in 21b9edf.

**1. The guard is itself hand-maintained** — fixed. `FrontendEnumContractTest` now derives the enum-shaped unions from the frontend source and fails when one is not registered, so the registration cannot fall behind silently. Verified the check can actually fail: unregistering `Frequency` turns the suite red, restoring it turns it green — a guard that cannot fail was the objection, so it was worth proving.

**2. `PROTOCOL` retained by comment only** — fixed. The constant is now `@Deprecated`, so reaching for it warns at the use site. Confirmed no `switch` over `UpgradeType` exists, so nothing warns spuriously.

**On the scope boundary** — the 500-on-unbindable-body defect is now tracked as #22 rather than living only in this description.

The nit about `PROTOCOL` and `EXPERIMENT` sharing a grey badge is left as-is, deliberately: it only matters if `PROTOCOL` becomes common, and it should not.

`mvn test` (9 contract tests among them), `npm run build` and eslint all pass.

Issue #22, opened 2026-08-18
fix: an unbindable request body returns 500 instead of 400


NaimElijah:
Split out of the review on #21, which found this while fixing the enum drift. Not fixed there because
it is a wire-contract change with a wider blast radius than that PR.

## Verified

A request body that cannot be bound returns **500**, not 400. Probed with a standalone MockMvc request
against `UpgradeController` and `GlobalExceptionHandler`:

```
POST /api/upgrades  {"title":"x","type":"NOT_A_TYPE"}
→ 500 {"status":500,"message":"Internal server error","path":"/api/upgrades",...}
```

## Cause

Jackson throws `HttpMessageNotReadableException` for an unbindable body — malformed JSON, an unknown
enum constant, a string where a number is expected. Spring's `DefaultHandlerExceptionResolver` maps
that to 400, but `ExceptionHandlerExceptionResolver` runs first, and `GlobalExceptionHandler` declares
a catch-all:

```java
@ExceptionHandler(Exception.class)
public ResponseEntity<ErrorResponse> handleGeneral(Exception ex, HttpServletRequest req) { ... 500 ... }
```

so the catch-all wins and the framework's own 400 never applies.

## Impact

Any client sending a malformed body is told the server broke, rather than that the request was wrong.
It is also logged as an unhandled failure, so genuine 500s are harder to spot among the noise. #19 has
removed the way users were reaching this — the five rejected upgrade types — but the sharp edge
remains for every other unbindable body.

Related: `handleGeneral` currently logs nothing at all. An unexpected 500 leaves no server-side trace
beyond the access log, which is worth fixing at the same time.

## Proposed fix

- Add `@ExceptionHandler(HttpMessageNotReadableException.class)` returning **400**, with a message that
  does not echo the parser's detail back to the client.
- Log in `handleGeneral` at ERROR with the exception, so a real 500 is diagnosable.
- Extend `GlobalExceptionHandlerTest`, which already pins the rest of the exception-to-status contract.

Changing a status from 500 to 400 is a wire-contract change. It is the correct status and no known
client branches on it, but it belongs in its own PR rather than riding along with an unrelated fix.

## Acceptance criteria

- [ ] Malformed JSON and unknown enum values return 400
- [ ] The response body carries no parser internals
- [ ] An unexpected exception is logged with its stack trace before the 500 is returned
- [ ] `GlobalExceptionHandlerTest` covers both

PR #24, opened 2026-08-18
docs: complete the rename to UpHealther and add the requirements document

NaimElijah:
## Problem

Two things, both consequences of the project being renamed and of a directory that was created but
never filled.

**1. The rename to UpHealther was half-finished.** The README heading and the licence title had been
updated; nothing else had. The old name was still on:

- the browser tab title, the navbar brand and the login heading — the three places a user actually
  sees the product named;
- **`LICENSE` §7**, which reserved the name `"HealthUpgrades"`. The trademark clause was protecting a
  name the project no longer uses, and saying nothing about the one it does;
- the README disclaimer, resume summary and licence section, plus a clone instruction pointing at a
  directory called `Health_Upgrades_Tracker`;
- the opening sentence of the architecture document.

**2. `docs/requirements/` held nothing but a `.gitkeep`.** The repository baseline requires a
requirements document, and there was none — so what the system must do existed only as code, tests and
inference.

## Approach

**For the rename, I drew a line between the project's name and identifiers that happen to contain the
old string,** and fixed only the first. Renamed: prose, documentation, and every user-visible surface.
Deliberately not renamed:

| Left alone | Why |
|---|---|
| `com.healthupgrades` Java package (162 files) | A mechanical but repository-wide refactor with no user-visible effect |
| `healthupgrades` database name, user and password | Baked into existing local volumes and `.env` files; changing the default breaks every existing environment |
| `demo@healthupgrades.com` | Seeded by Flyway `V2`, which is immutable — changing it needs a new migration |
| `healthupgrades-db` / `-backend` container names, Maven artifactId, `package.json` name | Build and infrastructure identifiers, not the product name |

None of those is "the project name being wrong". Each is a code or infrastructure change with its own
blast radius, and they belong in their own PR — happy to do it, but not folded into a documentation
change.

**For the requirements document, I recorded what the project meets today**, as asked: 33 functional
requirements, 15 business rules and invariants, 15 non-functional requirements, 4 non-goals and 6 open
questions. Every entry names the class or test that enforces it, so a claim in the document can be
checked rather than trusted — `BR-5` points at `UpgradeSchedulingServiceTest`, `NFR-9` at
`HexagonalArchitectureTest`, and so on.

It states *what*, points at `architecture.md` for *how* and the ADRs for *why*, and closes by saying
the project and its requirements are both still growing.

## Trade-off accepted

The document describes **implemented** behaviour rather than independently-established intent. That is
what was asked for, and it is the honest thing to say about a document derived from a finished
codebase — but it means anything the code gets wrong, the requirements would inherit. The open
questions section names the six things the code cannot answer, including the two known gaps (no
frontend test runner, no backend dependency audit) and the one open defect (#22), rather than
presenting the project as complete.

Priorities and delivery order are not recorded, because everything listed is already built and nothing
has needed ranking.

## How it was verified

- `mvn test` — passes.
- `npm run build` and `npx eslint src` — pass, clean.
- Every factual claim in the requirements document was checked against the code before writing it,
  including the ones easiest to get wrong: the `uq_progress_upgrade_date` unique constraint actually
  exists in `V1` (BR-6), the token lifetime really is 24 hours and non-refreshable (NFR-4), and the
  frontend really does offer exactly eight upgrade kinds (FR-9).
- Confirmed no occurrence of the old name survives in prose or UI; the remaining hits are all the
  identifiers listed above.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -7,6 +7,14 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 A "health upgrade" is the core domain object — a planned health improvement (habit, one-time action,
 experiment, goal, etc.) that moves through a lifecycle.
 
+## Requirements
+
+`docs/requirements/requirements.md` states what the system must do: capabilities, business rules and
+invariants, non-functional requirements, and the non-goals. Each entry names the class or test that
+enforces it, so it can be checked rather than trusted. Read it before proposing work — a change that
+satisfies no requirement, or contradicts one, is a conversation before it is code — and update it in
+the same commit when what the system must do changes.
+
 ## Architecture
 
 `docs/architecture/architecture.md` describes the system as it is: components, communication, data
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-HealthUpgrades — Proprietary Source-Available License
+UpHealther — Proprietary Source-Available License
 Version 1.0
 
 Copyright (c) 2026 Naim Elijah. All rights reserved.
@@ -113,7 +113,7 @@ SUMMARY (not part of the licence terms)
 7. NAME AND ATTRIBUTION
 
     This licence grants no right to use the name of the Author, the name
-    "HealthUpgrades", or any associated branding, whether to identify the origin
+    "UpHealther", or any associated branding, whether to identify the origin
     of a work, to endorse or promote anything, or otherwise.
 
     Authorship of the Software is asserted by the Author and is recorded in the
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# HealthUpgrades 🌱
+# UpHealther 🌱
 
 A full-stack health upgrade planning and tracking platform. Plan, activate, and track your health improvements — from drinking more water to meditating daily, from better sleep habits to product replacements.
 
@@ -54,10 +54,12 @@ A **Health Upgrade** is more than a habit. It can be:
 
 ## Architecture
 
-[`docs/architecture/architecture.md`](docs/architecture/architecture.md) describes the system as it
-actually is — components, how they communicate, the path a request takes end to end, external
-dependencies, and the structural decisions that are not visible from the file layout. Start there; the
-summary below is the short version.
+[`docs/requirements/requirements.md`](docs/requirements/requirements.md) states **what** the system
+must do — the capabilities, the business rules, and the non-goals — with the enforcing class or test
+named against each. [`docs/architecture/architecture.md`](docs/architecture/architecture.md) describes
+**how** it does it: components, how they communicate, the path a request takes end to end, external
+dependencies, and the structural decisions that are not visible from the file layout. Start with
+either; the summary below is the short version.
 
 ### Backend — Hexagonal (Ports & Adapters) + DDD
 
@@ -193,7 +195,7 @@ Events are published through a `DomainEventPublisher` outbound port (a Spring-ba
 1. **Clone the repository**
    ```bash
    git clone <repo-url>
-   cd Health_Upgrades_Tracker
+   cd UpHealther
    ```
 
 2. **Set up environment**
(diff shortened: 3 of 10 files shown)


NaimElijah (commented):
## Review

Self-review, so a checklist rather than a second opinion. Three things to fix — all consistency gaps
the change itself created — and several checks that came back clean.

---

### 1. The module guides point at the architecture document but not the requirements one

`backend/CLAUDE.md` and `frontend/CLAUDE.md` each open by pointing at
`../docs/architecture/architecture.md`. Neither mentions `requirements.md`, even though
`.claude/CLAUDE.md` and the README now both do.

That is the wrong half to leave out. Someone working inside `backend/` reads `backend/CLAUDE.md`, and
this PR has just added a document that says what their change is supposed to satisfy. A document
nobody is pointed at may as well not exist — which was the state `docs/requirements/` was in before
this PR.

### 2. `architecture.md`'s "Where the reasoning lives" table has no requirements row

That table exists so a reader who wants *why*, or conventions, or how to run the thing, is sent to the
right file. It now covers ADR-001, ADR-002, both module guides and the README — and not the one
document that answers "what is this system supposed to do".

The requirements document points at the architecture document. The link should run both ways, or the
two will drift apart in exactly the way the whole documentation set exists to prevent.

### 3. Nit — `#22` does not autolink in a repository file

`requirements.md` refers to "issue #22" in the open questions. GitHub autolinks bare `#N` references
in issues, pull requests and commit messages, but **not** in a rendered file blob, so it is plain text
a reader cannot follow. Small, and worth a full URL since the whole point of that section is to send
someone somewhere.

**Verdict: fix, trivial.**

---

### Checked, no change needed

**`README.md:155` — `### Health Upgrades` is correctly left alone.** It sits between `### Health Areas`
and `### Progress & Reflections` in the API overview, so it names the resource, not the product. The
domain concept is still "a health upgrade" and renaming it would have broken the vocabulary the whole
codebase, the aggregate `HealthUpgrade`, and `.claude/CLAUDE.md` share.

**No other occurrence of the old name survives in prose or UI.** The remaining hits are the identifiers
the PR description lists as deliberately untouched, all of which contain `com.healthupgrades` or a
lowercase `healthupgrades` database/container name.

**Every relative link in the new document resolves** — `../architecture/architecture.md`, `../ADRs/`
and the ADR-003 link were each checked against the filesystem, not assumed.

**The counts in the PR description match the document**: FR-1 to FR-33, BR-1 to BR-15, NFR-1 to
NFR-15, four non-goals, six open questions.

**The factual claims easiest to get wrong were verified against the code, not from memory** — the
`uq_progress_upgrade_date` unique constraint really is declared in `V1` (BR-6, and a claim I had made
in a doc comment previously without checking), the token lifetime is 86400000ms and non-refreshable
(NFR-4), and the frontend offers exactly eight upgrade kinds (FR-9).

---

### On the boundary I drew

Splitting "the project's name" from "identifiers containing the old string" is the right call, but it
is a judgement, not a fact, and it leaves the repository in a genuinely mixed state: the product is
UpHealther everywhere a person looks, and HealthUpgrades everywhere a machine looks. That is defensible
— the database name and the Java package are not the product's name — but it should be a decision
someone made on purpose, which is what the PR description is for, rather than something discovered
later and mistaken for an oversight.

NaimElijah:
All three review findings addressed in 2b66d7c.

**1. Module guides didn't point at the requirements document** — fixed. `backend/CLAUDE.md` and `frontend/CLAUDE.md` now open with it alongside the architecture pointer, each framed for its own module: the backend guide names the business rules and their enforcing tests, the frontend guide names the rules the UI is expected to respect.

**2. The architecture table had no requirements row** — fixed. "What is the system supposed to do, and what is it deliberately not doing?" now heads that table, so the link runs both ways instead of only from requirements to architecture. Verified the relative path resolves from `docs/architecture/`.

**3. Bare `#22` doesn't autolink in a file blob** — fixed, now a full URL.

Nothing else changed; the rename and the requirements document are as reviewed.

PR #27, opened 2026-08-18
docs: add arch-diagrams — seven views of the architecture, three generated from source

NaimElijah:
## Problem

`architecture.md` described the system in prose with four diagrams scattered through it. There was no
single place to go and *see* the shape — and the context map in it was hand-drawn, so it recorded what
someone believed the dependency graph to be rather than what it was.

## Approach

`docs/architecture/arch-diagrams/README.md` is the page GitHub renders when you open the folder, so
landing there shows the whole architecture rendered rather than a list of filenames to click through.
That is the reason it is one page and not seven files.

Seven diagrams, ordered outside in:

| # | Diagram | Source |
|---|---|---|
| 1 | System context | Hand-written |
| 2 | Bounded-context map | **Generated** from the imports |
| 3 | How to read a context — the hexagon key | Hand-written |
| 4 | Inside `upgrade` — class diagram | **Generated** from the package layout |
| 5 | The upgrade lifecycle — state machine | Hand-written |
| 6 | Request lifecycle — sequence | Hand-written |
| 7 | Data model — ER | **Generated** from the Flyway migrations |

**The split is not arbitrary.** Three of these restate something the code already states, and would be
wrong on the next import, field or column — those are generated by `generate.py` into marked regions,
and CI fails when the committed output stops matching the source. The other three encode intent, which
no generator can derive and which does not rot.

Diagram 7 did not exist anywhere before: the schema had no picture at all.

## Trade-off accepted

**The four diagrams moved out of `architecture.md` rather than being copied,** so there is exactly one
source for each and the document links to them at the point each used to sit. The cost is that
`architecture.md` is now prose-only and a reader has to follow a link to see a picture. Copying was
the worse option: the context map in the folder is generated and the copy left behind would not be, so
they would diverge — which is the failure this whole change exists to prevent.

**The class diagram covers `upgrade` only, not all ten contexts.** One worked example plus the "how to
read a context" key teaches the shape well enough to read the other nine from the source. Ten
near-identical diagrams is scrolling, not understanding. If they turn out to be wanted, they belong in
a `per-context/` subfolder, linked but out of the reader's way.

**`generate.py` parses Java with regular expressions rather than a real parser.** That is adequate for
this codebase, which is uniform, and it keeps the tool dependency-free — but it is not a general Java
parser and would need care if the source got more exotic. Two bugs of exactly that kind were found and
fixed while building it, both listed below.

## How it was verified

- **The drift check can actually fail.** Adding a class to the `upgrade` context turns `--check` red;
  removing it turns it green. A guard that cannot fail is worse than no guard, so this was checked
  rather than assumed.
- Two generator bugs found by reading the output against the source, not by trusting it:
  - `UpgradeJpaRepository --|> HealthUpgrade : extends` — false. It extends
    `JpaRepository<HealthUpgrade, UUID>`; the type *argument* was being read as a supertype.
  - `progress_entries` showed `upgrade_id UK` and `date UK` as separate unique columns. The constraint
    is composite — `UNIQUE (upgrade_id, date)` — and claiming each is independently unique is a
    different, false statement. Mermaid has no composite notation, so single-column uniques are marked
    and composite ones are described in the prose beside the diagram.
  - Port packages were stamped `<<outbound port>>` for both the interface and the records passed across
    it; they are now distinguished as `<<outbound port>>`, `<<port record>>` and `<<use-case record>>`.
- `mvn test` passes; no production code changed.

## Note on what I could not verify

I have **not seen these rendered**. Mermaid syntax here is deliberately conservative and auto-laid-out,
so the risk is low, but if any block fails to parse it will show as an error box on GitHub rather than
a diagram. Worth a glance at the Files tab before merging — that is the one check I cannot do from
here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,19 @@ jobs:
       # its database without an NVD API key and fails outright without one. See the known gaps in
       # ADR-002 for what enabling it requires.
 
+  diagrams:
+    name: Diagrams - Drift check
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      # Three of the diagrams in docs/architecture/arch-diagrams/ are derived from the imports, the
+      # package layout and the migrations. Regenerating here and failing on a difference is what stops
+      # them describing an architecture the code no longer has — the failure ADR-002 was written about.
+      - name: Check the generated diagrams still match the source
+        run: python3 docs/architecture/arch-diagrams/generate.py --check
+
   frontend:
     name: Frontend - Build & Lint
     runs-on: ubuntu-latest
--- a/docs/architecture/arch-diagrams/README.md
+++ b/docs/architecture/arch-diagrams/README.md
@@ -150,8 +150,11 @@ Relations shown are **implements**, **extends**, and *holds a field of this type
 that reveals the wiring. Accessors and framework plumbing are left out on purpose; the interesting
 content is which layer a type sits in and which way its arrows point.
 
-Scoped to this context's own types, so dependencies on `common` and on Spring do not appear —
-`UpgradeService` also holds a `DomainEventPublisher` and a `Clock`, neither of which is drawn here.
+Two things are therefore absent by construction, and neither means what it looks like. Dependencies on
+`common` and on Spring are out of scope — `UpgradeService` also holds a `DomainEventPublisher` and a
+`Clock`. And a type *constructed in a method* rather than held as a field draws no arrow, which is why
+nothing connects `UpgradeBeansConfig` to `UpgradeSchedulingService` even though wiring that bean is the
+only reason the class exists.
 
 <!-- generated:class-diagram -->
 ```mermaid
@@ -208,17 +211,29 @@ classDiagram
         <<outbound port>>
     }
 
-    %% ---- application ----
-    class UpgradeBeansConfig {
-        <<application>>
+    %% ---- port record ----
+    class UpgradeTrackingSummary {
+        <<port record>>
     }
+
+    %% ---- application ----
     class UpgradeService {
         <<application>>
+    }
+
+    %% ---- bean wiring ----
+    class UpgradeBeansConfig {
+        <<bean wiring>>
     }
 
     %% ---- inbound port ----
     class UpgradeQuery {
         <<inbound port>>
+    }
+
+    %% ---- use-case record ----
+    class UpgradeDetails {
+        <<use-case record>>
     }
 
     %% ---- web adapter ----
@@ -258,16 +273,6 @@ classDiagram
     }
     class UpgradeRepositoryAdapter {
         <<persistence adapter>>
-    }
-
-    %% ---- port record ----
-    class UpgradeTrackingSummary {
-        <<port record>>
-    }
-
-    %% ---- use-case record ----
-    class UpgradeDetails {
-        <<use-case record>>
     }
 
     %% ---- relations: implements, extends, and fields typed by another type ----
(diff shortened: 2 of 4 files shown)


NaimElijah (commented):
## Review

Self-review. Two things to fix, one gap to name, and one correction to the PR description.

---

### Correction to the description first

The PR says I had not seen these rendered and asked you to check the Files tab. That is now out of
date: I validated all seven blocks against the **real Mermaid grammar** (the `mermaid` package driven
over jsdom in a scratch directory, not a lookalike), and every one parses:

```
  OK  #1  flowchart LR        OK  #5  stateDiagram-v2
  OK  #2  flowchart TD        OK  #6  sequenceDiagram
  OK  #3  flowchart LR        OK  #7  erDiagram
  OK  #4  classDiagram
```

That closes the risk I actually cared about — the multi-word `<<domain model · enum>>` stereotypes,
which I was not certain the class-diagram grammar accepted. It parses. What this still does not prove
is that the *layout* is legible once rendered, only that nothing errors.

---

### 1. The class diagram silently omits method-level wiring

The README says relations are "implements, extends, and *holds a field of this type*". That is
accurate, but it has a consequence it does not warn about: `UpgradeBeansConfig` constructs
`UpgradeSchedulingService` in a `@Bean` method rather than holding it as a field, so **no arrow is
drawn between them**.

A reader who knows the domain service is hand-wired will look for that edge, not find it, and
reasonably conclude either that the wiring is missing or that the diagram is wrong. Neither is true.
One clause in the prose fixes it.

### 2. `UpgradeBeansConfig` is stereotyped as if it were a use case

It carries `<<application>>`, the same stamp as `UpgradeService`. One is a transactional use case, the
other is Spring configuration that exists only because the domain services deliberately carry no
stereotype annotation. Flattening that distinction loses the very thing the layout is trying to show.

---

### Gap worth naming, not fixing here

**There is no frontend diagram.** All seven views are backend-centric — the SPA appears as a single
box in §1 and never again. The frontend has real structure worth drawing (the single axios instance
and its two interceptors, the two providers, the query cache, the route guard), and none of it is
visible here.

That is a defensible scope line for this PR and a genuine hole in the set. It should be its own change
rather than an eighth diagram bolted on, and it would want the same treatment: generated where
mechanical, hand-drawn where conceptual.

---

### Checked, no change needed

**The drift check is newline-agnostic.** `generate.py` writes CRLF on Windows and LF on Linux, which
looked like a guaranteed CI failure. It is not: Python reads with universal newlines, so both sides of
the comparison are normalised before it happens. Confirmed empirically — the job passes on CI in 4s
having been generated on Windows.

**The generated content was read against the source rather than trusted.** Three defects came out of
that and are already fixed in this branch: a false `extends` edge from a generic type argument, a
composite `UNIQUE` rendered as two independent single-column keys, and port packages stamping their
interface and their records identically.

**No production code changed** — `mvn test` passes, and the only non-docs file touched is `ci.yml`.

NaimElijah:
Both review findings fixed in the latest commit.

**1. The class diagram silently omitted method-level wiring** — the prose now states what is absent by construction: dependencies on `common` and Spring are out of scope, and a type constructed in a method rather than held as a field draws no arrow, which is why nothing connects `UpgradeBeansConfig` to `UpgradeSchedulingService`.

**2. `UpgradeBeansConfig` stamped as a use case** — now `<<bean wiring>>`, so the difference between a transactional use case and the configuration that hand-wires the framework-free domain services is visible rather than flattened.

Regenerated and revalidated: all seven blocks still parse against the real Mermaid grammar.

The frontend-diagram gap I raised is deliberately left for its own change.

Issue #36, opened 2026-08-18
Implementing Dark mode Features & Requirements


NaimElijah:
## Why this is more tractable here than usual

The UI is already centralised in a way that makes a theme pass realistic:

- `src/components/ui/` holds the primitives (`Button`, `Card`, `Input`, `Select`, `Badge`, `Modal`,
  `EmptyState`, `PageHeader`) and each keeps its colour choices in a single lookup map.
- Category and status colours are similarly centralised — `notificationMeta.ts`, `statusMap` in
  `UpgradeStatusBadge`, `typeMap` in `UpgradeTypeBadge`, `difficultyVariant` in `UpgradeCard`.
- Tailwind is already the styling system, so `darkMode: 'class'` plus `dark:` variants is a supported
  path rather than a rewrite.

## Where the work actually is

Not in the primitives — in the **pages**, which carry a lot of literal `bg-white`, `text-gray-900`,
`bg-gray-50` utilities inline. That is the bulk of the diff and the part most likely to be missed in
patches.

## Suggestions

1. **Introduce semantic tokens before touching pages.** `bg-surface`, `text-muted`, `border-subtle` —
   defined once in the Tailwind config for both themes. Otherwise every element needs a `dark:` twin and
   the next theme change costs the same again.
2. **Set `darkMode: 'class'` and drive it from a provider** that reads `localStorage` and falls back to
   `prefers-color-scheme`, following the pattern `AuthProvider` and `NotificationProvider` already
   establish.
3. **Apply the class before first paint** — a small inline script in `index.html` — or the app flashes
   light on every load for dark-mode users.
4. **Do the primitives and one page first** to prove the token set, then sweep. The colour maps listed
   above are where most of the value is concentrated.
5. **Check contrast rather than eyeballing it.** The existing palette leans on `text-gray-400` and
   `text-gray-500` for secondary text, which is already borderline in light mode and worse inverted.

## Open questions

- Three-state toggle (light / dark / system) or two?
- Does the emoji-based iconography still read on a dark surface? Several are effectively unstyled.

---

*Opened from the **Future Improvements** list in the README. Scope, priority and whether to do this at
all are open — that is what the brainstorm is for. When it is settled, the outcome belongs in
[`docs/requirements/requirements.md`](../blob/dev/docs/requirements/requirements.md), which currently
records only what the project already meets.*

PR #38, opened 2026-08-19
feat(frontend): dark mode via semantic colour tokens

NaimElijah:
## Problem

The frontend had no theming layer at all: `tailwind.config.js` was bare, `src/index.css` held only the
three `@tailwind` directives, and there was not one `dark:` variant, `prefers-color-scheme` query or
theme provider anywhere in `src/`. Colour was written as literal Tailwind palette utilities at 247 call
sites across 29 files.

Settles the brainstorm opened in #36. Ticks the `Dark mode` box in the README's future-improvements list.

The audit that preceded this found the existing palette already carried WCAG failures **in light mode**,
before dark mode was considered: `text-gray-400` is **2.54:1** on white, and `border-gray-300` on form
inputs is **1.47:1**. A mechanical `dark:` sweep would have preserved both forever.

## Approach

Every colour becomes a **semantic token** — a role, not a shade. 55 tokens declared in
`tailwind.config.js` as `rgb(var(--color-x) / <alpha-value>)`, given their light and dark values in
`index.css` under `:root` and `.dark`.

Channel triplets rather than hex are forced by two call sites that use Tailwind's opacity modifier on a
colour that must theme — the modal scrim and the unread-notification tint. A hex-valued CSS variable
cannot take that modifier; a channel triplet can.

`darkMode: 'class'`, not `'selector'`: `package.json` declares `^3.4.0` and `'selector'` only landed in
3.4.1, so a fresh install resolving 3.4.0 exactly would break.

Theme state is three-valued — light / dark / **system**, where system follows the OS live rather than
only at boot. Persistence is `localStorage` only: no backend change, no migration, no endpoint.

A classic (non-module) inline script in `<head>` applies the `dark` class before first paint. It must
stay non-module — a `type="module"` inline script gets bundled and deferred, which reintroduces the flash.

## Trade-offs accepted

- **Light mode visibly changes.** Meeting AA moves form-control borders (1.47:1 → 3.73:1), the faintest
  text (2.54:1 → 4.83:1), the progress fill and the unread-count badge. Inputs now read as noticeably
  more outlined. The light text ramp compresses from 17.7/10.3/6.0/2.5 to 17.7/10.3/6.0/4.8 — hierarchy
  survives, the two faintest steps sit 1.24x apart instead of 1.9x. Unavoidable at AA.
- **One deliberate AA exception, stated rather than hidden.** `line` (1.24:1) and `line-subtle` (1.13:1)
  do not meet 3:1. WCAG 1.4.11 covers visual information required to *identify components and states*; a
  card's decorative outline and a list's divider identify nothing, and the card is not a control. Control
  boundaries — `Input`, `Select`, the modal panel, the focus ring — *are* enforced at 3:1 via
  `line-strong`. Changing `line` to `#7C8593` gets 3:1 everywhere and moves nothing else.
- **`system` is the default, so this ships as a visible change for existing users.** Anyone on a dark OS
  gets dark on the next deploy without opting in. Correct semantics, but it is not a silent release.
- **Emoji are the known, unfixable limitation.** Sidebar nav and empty-state glyphs carry their own
  colours and will read as bright specks on dark surfaces. No CSS remedy; `ThemeToggle`'s inline SVG
  establishes the pattern for a follow-up icon set.
- **Tailwind fails silently** — `bg-surfce` emits no CSS and no error. Mitigated by `npm run check:colours`,
  which greps `src/` for any surviving palette class and runs in CI.

## What was done

Twelve commits, each leaving lint, test and build green. The token sweep lands **before** the provider so
commits 5-8 review as "did the class names map correctly?" against a still-light app, and commit 9 reviews
as "does the switch work?" against a fully-swept one.

| # | Commit | Contents |
|---|---|---|
| 1 | `docs: record the frontend test harness decision (ADR-004)` | ADR-002 built a floor under the backend and left the frontend with none |
| 2 | `chore(frontend): add Vitest, jsdom and Testing Library` | Config, setup, CI step |
| 3 | `fix(frontend): define the fadeIn keyframes the toasts animate with` | Pre-existing bug: the arbitrary-value form is exactly what stopped Tailwind emitting the keyframes, so the toast animation had never run |
| 4 | `feat(frontend): define the semantic colour tokens` | Config + CSS. No consumers yet |
| 5-8 | the sweep | 29 files, in four reviewable slices: primitives, shell, notifications, upgrade views and pages |
| 9 | `feat(frontend): add the theme provider and the pre-paint boot script` | Dark mode becomes reachable here |
| 10 | `feat(frontend): add the three-state theme toggle` | Navbar and both auth pages |
| 11-12 | `.gitignore`, docs | Requirements, architecture, `frontend/CLAUDE.md`, README |

Two structural fixes taken while in the neighbourhood, both flagged in the plan: `Sidebar` moves to
`bg-surface` so the rail no longer collides with the page canvas, and the notification categories
separate (WARNING red→amber, INFO brand-blue→cyan, REMINDER orange→violet) so a category colour is no
longer indistinguishable from a severity signal.

## How it was verified

Automated, all of it in CI:

```
cd frontend && npm run lint            # eslint --max-warnings 0 — clean
cd frontend && npm run test            # vitest run — 31 passed
cd frontend && npm run build           # tsc && vite build — 190 modules, built
cd frontend && npm run check:colours   # "No palette colours in src/ — all 63 source files use tokens"
```

Contrast was **computed**, not eyeballed — every token pair checked against WCAG AA with a purpose-written
script. Worst text pair anywhere after the change: **4.63:1**. Worst non-text: **3.29:1**.

Two of the tests were confirmed to bite rather than merely pass: removing the `removeEventListener`
cleanup makes StrictMode register two `matchMedia` listeners and two tests fail. `check:colours` was
likewise confirmed by injecting `bg-white` and `border-t-blue-600`, both caught.

Manual pass in a real browser across login, register, dashboard, daily check-in and a modal, in both
themes: theme persists across reload, `color-scheme: dark` reaches the native controls, the focus ring
offset resolves to the dark surface rather than painting a white halo, and the modal panel is discernible
against the scrim.

**Not verified:** first paint was not frame-captured, so "no flash" rests on the script ordering plus
observed persistence, not on watching it.

## ADRs

- `docs/ADRs/ADR-004-frontend-test-harness.md` — new, records Vitest + jsdom + Testing Library and why
  Vitest 3 rather than 4 (Vitest 4 requires Vite >= 6; this project is on Vite 5.4.21).

## Out of scope, tracked separately

- `aria-pressed` on the four pill groups and `aria-invalid`/`aria-describedby` on `Input`/`Select`. Real
  bugs, equally broken in light mode today, not theming bugs.
- Replacing emoji UI icons with an SVG set.
- Pre-existing, unrelated: health-area cards render `water_drop` / `fitness_center` / `bedtime` as literal
  text — `V2__seed_data.sql` stores Material Symbols names while `HealthAreasPage.tsx:91` renders that
  field as an emoji. Identical in light mode.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,11 @@ jobs:
       - name: Lint frontend
         run: cd frontend && npm run lint
 
+      # Tailwind emits nothing for an unknown class and raises no error, so a mistyped or missed colour
+      # renders as no colour at all rather than failing. This is the only check that catches it.
+      - name: Check no palette colours bypass the theme tokens
+        run: cd frontend && npm run check:colours
+
       # After the lint, so a failure here is a behaviour failure rather than a style one. `npm run test`
       # is `vitest run` and not the watcher — a bare `vitest` would hold the runner until the job times
       # out. See ADR-004 for why the frontend has a test harness at all.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,12 @@ jobs:
       - name: Lint frontend
         run: cd frontend && npm run lint
 
+      # After the lint, so a failure here is a behaviour failure rather than a style one. `npm run test`
+      # is `vitest run` and not the watcher — a bare `vitest` would hold the runner until the job times
+      # out. See ADR-004 for why the frontend has a test harness at all.
+      - name: Test frontend
+        run: cd frontend && npm run test
+
       # Scoped to what ships. Auditing the dev tree here would redden every unrelated PR whenever an
       # advisory lands against the Vite or ESLint toolchain, none of which reaches a user's browser.
       - name: Audit shipped frontend dependencies for known vulnerabilities
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ backend/.mvn/wrapper/maven-wrapper.jar
 frontend/node_modules/
 frontend/dist/
 frontend/.vite/
+# Vitest and npx write a cache here when a command is run from the repo root by mistake.
+node_modules/
 
 # Environment
 .env
--- a/README.md
+++ b/README.md
@@ -269,11 +269,19 @@ mvn verify    # the above plus integration tests — needs a running PostgreSQL
 bean graph or a missing Flyway migration — neither of which a unit test can see. Start a database first
 with `docker-compose up -d postgres`.
 
-### Frontend Build (includes type check)
+### Frontend Tests
 ```bash
 cd frontend
-npm run build
+npm run test           # Vitest + jsdom + Testing Library
+npm run build          # type check and production build
+npm run lint           # ESLint, zero warnings tolerated
+npm run check:colours  # fails on any Tailwind palette colour used outside the theme tokens
 ```
+
+`npm run test` covers the theme provider — that an explicit choice beats the operating system, that
+`system` follows it live, that the media-query listener is cleaned up under `React.StrictMode`, and that
+a browser with storage blocked still boots. See
+[ADR-004](docs/ADRs/ADR-004-frontend-test-harness.md) for why the harness exists and what it cannot do.
 
 ## Environment Variables
 
@@ -297,7 +305,7 @@ npm run build
 - [ ] Gamification: achievements, badges, levels
 - [ ] Integration with wearables (steps, sleep data)
 - [ ] Weekly/monthly progress email reports
-- [ ] Dark mode
+- [x] Dark mode
 - [ ] Internationalization (i18n)
 
 ## Screenshots
--- a/docs/ADRs/ADR-004-frontend-test-harness.md
+++ b/docs/ADRs/ADR-004-frontend-test-harness.md
@@ -1,0 +1,134 @@
+# ADR-004: Test the frontend with Vitest, jsdom and Testing Library
+
+- **Status:** Accepted
+- **Date:** 2026-08-19
+- **Scope:** `frontend/` only. No effect on the backend, the runtime, or anything that ships to a
+  browser — every package added here is a development dependency.
+
+## Context
+
+[ADR-002](ADR-002-close-the-gap-between-the-described-and-enforced-architecture.md) built a floor under
+the backend: an integration test that boots the application against a real PostgreSQL, ArchUnit rules
+that fail on a layering violation, and tests for the two services that had none. It left the frontend
+without an equivalent, and `docs/requirements/requirements.md` §6 has recorded that gap as an open
+question ever since:
+
+> **The frontend has no test runner.** No vitest, no jest. Frontend logic is currently verified only by
+> the type checker, the linter and inspection. Introducing one is a dependency decision that deserves
+> an ADR.
+
+This is that record.
+
+Until now the gap was tolerable because the frontend held almost no logic worth testing. Its
+components render server state that TanStack Query fetches, and the one piece of real behaviour —
+that the mirrored enums cannot drift from the backend's — is already enforced from the backend by
+`FrontendEnumContractTest`. Type checking and a zero-warning lint genuinely covered the rest.
+
+Dark mode changes that. `ThemeProvider` holds branching logic that no type can express and no reviewer
+can reliably eyeball:
+
+- an explicit choice must beat the operating system's preference, but only when it is explicit;
+- `system` must follow the OS **while the tab is open**, which means a `matchMedia` subscription with a
+  cleanup that `React.StrictMode` will exercise twice on every mount;
+- reading `localStorage` throws outright where site data is blocked, and there is no error boundary
+  above this provider, so an unguarded read blanks the whole application.
+
+Each of those is a small, pure, deterministic assertion. Each is also the kind of defect that ships
+green and surfaces months later on somebody else's machine.
+
+## Decision
+
+Add **Vitest** with **jsdom** and **Testing Library** as development dependencies, and run them in CI
+between the lint and the audit.
+
+Four packages, all `devDependencies`: `vitest`, `jsdom`, `@testing-library/react`,
+`@testing-library/dom`. Nothing reaches a user's browser, so `npm audit --omit=dev` — the audit scope
+ADR-002 chose deliberately — is unaffected.
+
+Four choices inside that decision are worth stating, because each has a failure mode:
+
+- **Vitest is pinned to `^3.2.7`, not `^4`.** Vitest 4 declares a peer dependency on Vite `^6 || ^7 ||
+  ^8`; this project is on Vite 5. Taking Vitest 4 would mean a Vite major upgrade riding along inside a
+  styling change. Vitest 3 depends on Vite `^5.0.0 || ^6.0.0 || ^7.0.0-0` and needs nothing moved.
+- **Configuration lives in `vitest.config.ts`, merged over `vite.config.ts` with `mergeConfig`.** A
+  `test` key inside `vite.config.ts` would make the production build configuration type-depend on a
+  development-only package. Keeping them separate without the merge is worse: Vitest reads
+  `vitest.config.ts` *instead of* `vite.config.ts` when both exist, so the React plugin and the dev
+  proxy would silently stop applying to tests. The merge is what keeps one source of truth.
+- **No globals.** `describe`, `it` and `expect` are imported in each test file rather than injected
+  ambiently. This leaves `tsconfig.json`'s `types` array — and therefore the production type check —
+  untouched, at the cost of one import line per file.
+- **Tests are colocated under `src/`, not in a separate tree.** `tsconfig.json` includes `src`, so
+  `tsc` type-checks the tests as part of `npm run build`. A test that no longer compiles is a broken
+  test and should redden the build.
+
+## Consequences
+
+**What this makes easy**
+
+- Frontend behaviour can be asserted rather than inspected. `requirements.md` entries covering the SPA
+  can now name a real test in their "Enforced by" column instead of naming a component and asking the
+  reader to trust it.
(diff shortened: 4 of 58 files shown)


NaimElijah (commented):
## Review

Reviewed the full `dev...feature/dark-mode` diff — 57 files, 2676 insertions. The token architecture
holds up: I cross-checked all 55 tokens declared in `tailwind.config.js` against the `--color-*` values
in `index.css` and there are no orphans in either direction — every token has a value in `:root` *and*
in `.dark`, and nothing is defined in CSS that the config doesn't reference. That is the failure mode
this design is most exposed to, given Tailwind emits nothing and says nothing for an unknown class, and
it's clean.

Five findings below. One is a real accessibility defect, one is a missing test that guards the branch's
central invariant, and three are smaller. Verified against the built CSS or the source rather than
inferred.

---

### 1. The `prefers-reduced-motion` rule is inert — `src/index.css`

The override is declared inside `@layer base`:

```css
@media (prefers-reduced-motion: reduce) {
  .animate-fade-in { animation: none; }
}
```

Tailwind emits `base` before `utilities`, and both selectors have specificity (0,1,0), so the utility
wins on source order. Confirmed by byte offset in `dist/assets/index-*.css`:

```
8455  .animate-fade-in{animation:none}                    <- base layer
10991 .animate-fade-in{animation:fade-in .2s ease-out}    <- utilities layer, wins
```

Someone who has asked their OS for reduced motion still gets the animation. The comment above the rule
states the intent plainly, so this reads as working and doesn't.

Fix: move the media query out of `@layer base` entirely. Unlayered CSS in the same stylesheet beats
Tailwind's layers, so declaring it after the `@tailwind` directives is enough and needs no `!important`.

### 2. Nothing binds the boot script to `ThemeProvider` — `index.html`, `src/contexts/ThemeProvider.tsx`

The theme resolution rule is implemented twice — once in the inline boot script and once in the
provider — in two languages, and they must agree on the storage key (`theme`), the three accepted
values, and the precedence of an explicit choice over the OS preference. Three separate comments assert
they are "kept in step". Nothing enforces it.

Change `STORAGE_KEY` in the provider and all 31 tests still pass while the flash-of-wrong-theme comes
back silently — which is the one defect this whole mechanism exists to prevent, and the one no reviewer
will catch by reading.

It is testable: read `index.html`, extract the script body, run it in jsdom against a seeded
`localStorage` and `matchMedia`, and assert the resulting class matches what `ThemeProvider` resolves to
from the same inputs. That turns "kept in step" from a comment into a contract.

### 3. `peer-hover:text-fg` never fires — `src/components/theme/ThemeToggle.tsx`

Emitted as:

```css
.peer:hover~.peer-hover\:text-fg { color: rgb(var(--color-fg) / …) }
```

The general sibling combinator requires the **input** to be hovered, and the input carries `sr-only` —
`position:absolute; width:1px; height:1px; clip:rect(0,0,0,0)`, sitting behind the icon it labels. The
pointer lands on the `<span>`, never on the 1px input. The hover affordance on the two unselected icons
does not exist.

`peer-checked:` and `peer-focus-visible:` are unaffected — those are driven by the input's own state,
not by pointer position, so they work. It's specifically the hover that's dead.

Fix: add `group` to the `<label>` and use `group-hover:text-fg` on the span.

### 4. `check:colours` can't see `index.html` — `frontend/scripts/check-colours.mjs`

`sourceFiles()` walks `src/` and filters on `/\.tsx?$/`. But `index.html` is in Tailwind's `content`
glob, so a palette class written there compiles and ships while passing the guard clean. The file is
currently classless, so this is a gap rather than a live bug — but the guard's whole purpose is
catching what nothing else can, and this is a hole in exactly that.

Same category: the regexes match named utilities only, so arbitrary values (`bg-[#fff]`,
`text-[rgb(0,0,0)]`) pass through. Worth at least a comment stating the boundary.

### 5. The `difficultyVariant` type lies — `src/components/upgrade/upgradeMeta.ts`

```ts
export const difficultyVariant: Record<string, BadgeVariant> = { … }
```

Without `noUncheckedIndexedAccess`, TypeScript types `difficultyVariant[x]` as `BadgeVariant` — never
`undefined`. Both call sites correctly write `?? UNKNOWN_DIFFICULTY_VARIANT`, but to the compiler that
fallback is unreachable code. It works at runtime; the type just doesn't describe it.

`Partial<Record<UpgradeDifficulty, BadgeVariant>>` would make the compiler *require* the fallback rather
than merely tolerate it, and would catch a future call site that forgets it. Given the docstring says
this map exists precisely because the old inline ternary mishandled unknown values, having the type
enforce the fix seems worth the two words.

---

### Checked and clean

- All 55 tokens present in both themes; no config token without a value, no CSS value without a consumer.
- Both `difficultyVariant` call sites guard the lookup.
- `ringOffsetColor.DEFAULT` is set, so `ring-offset-2` doesn't paint a white halo on dark — the single
  easiest thing to get wrong here.
- `classList.toggle(name, force)` two-argument form and the `removeEventListener` cleanup make the
  provider StrictMode-safe, and there are tests asserting both.
- The `localStorage` reads and writes are both guarded, with tests covering the throwing case.
- CI ordering is sensible: lint, then `check:colours`, then tests, then the shipped-dependency audit.
- `npm run test` is `vitest run`, not the watcher — CI won't hang.
- Verified green locally: lint clean, 31/31 tests, build 190 modules, `check:colours` reports all 63
  source files on tokens.

NaimElijah:
## All five findings addressed

Five commits, one per finding. Each was verified by breaking it deliberately and watching the guard
fire, not just by watching the suite stay green.

| Finding | Commit | Verification |
|---|---|---|
| 1. Reduced-motion rule inert | `19c93ae` | Override now at byte 22503 vs the utility at 10969 — it follows and wins |
| 2. Boot script unbound to provider | `ef58450` | Renaming the storage key fails 3 cases; the other 32 tests stay green |
| 3. Dead hover on the toggle | `13fb0ca` | Emitted order: `group-hover` → `peer-checked` → `peer-checked:group-hover` |
| 4. `check:colours` blind to `index.html` | `2d0d6e5` | Injected `bg-white` into the shell; now reported as `index.html:34` |
| 5. `difficultyVariant` type | `cc68009` | Compiler now requires the map to be exhaustive |

Suite is 31 → 35 tests. Lint, build and `check:colours` all green; the guard now covers 65 files
rather than 63.

### Two departures from the review, with reasons

**Finding 5 — I did not do what the review suggested.** It proposed
`Partial<Record<Difficulty, BadgeVariant>>` to force the compiler to require the `??` fallback. That
is the wrong trade: `Difficulty` is a closed union of exactly the three keys the map already holds, so
`Partial` would buy a required fallback at the cost of no longer requiring the map to be exhaustive —
you could delete `HARD` and nothing would complain. Used `Record<Difficulty, BadgeVariant>` instead,
which keeps exhaustiveness, and documented why the runtime fallback is still not redundant: `difficulty`
arrives as unvalidated JSON, so its static type is a claim about the backend rather than a fact about
the value.

**Finding 3 — the first fix carried a wrong comment.** Swapping `peer-hover` for `group-hover` risks
the selected pill losing its inverted text on hover, so I added `peer-checked:group-hover:text-fg-inverted`
and wrote that it was rescuing the checked state. Checking the built CSS showed Tailwind already emits
`peer-checked` after `group-hover`, so the plain version would have been fine. Kept the explicit class —
that ordering is internal to Tailwind rather than a guarantee, and the failure mode would be
dark-on-blue text nobody thought to test — but corrected the comment to say it pins the outcome rather
than rescues it.

### One implementation note

The contract test first used `node:fs`, which broke `tsc`: the app's `tsconfig.json` has
`types: ["vite/client"]` and ADR-004 deliberately keeps Node types out of it. Switched to Vite's `?raw`
import, which needs no config change and resolves the file the same way the build does.

`frontend/CLAUDE.md` updated in the same commits: the boot-script/provider agreement is now enforced by
a named test rather than by three comments asking for care, and the colour guard's scope and its one
blind spot (arbitrary values like `bg-[#fff]`) are stated.

NaimElijah:
## CI caught a sixth defect the review missed

The frontend job failed on its **first ever real execution** with `TypeError: webidl.util.markAsUncloneable is not a function`, thrown from undici while jsdom loaded. Fixed in `847d1a9`; all three checks now pass.

Root cause was not the new test. `jsdom@30` declares `engines: ^22.22.2 || ^24.15.0 || >=26.0.0` and depends on `undici@8`, which requires Node `>=22.19.0`. CI and `frontend/Dockerfile` both run Node 20, so npm installed it under an `EBADENGINE` warning and it died at import. Everything passed locally because this machine runs Node 24 — the mismatch was invisible from a developer machine by construction.

Pinned to `jsdom@^29`, which accepts `^20.19.0 || ^22.13.0 || >=24.0.0` and covers both. Same principle that pinned Vitest to 3 rather than 4: the harness bends to the project's runtime, not the reverse.

Worth being blunt about how this got here. The CI test step was added in `95b184c` and merged without ever running — the job had no pull request to run against — so a plain unsupported-engine mismatch sat undetected until now. A guard that has never executed is not yet a guard, and neither the plan nor my review checked whether the declared engine range covered the CI runner.

### The larger issue, deliberately not fixed here

**Node 20 reached end of life in April 2026.** CI pins `node-version: '20'` and `frontend/Dockerfile` builds `FROM node:20-alpine`, so the project builds and ships on an unsupported runtime. That is the real reason jsdom 30 was unreachable, and pinning jsdom back treats the symptom.

Moving off it touches the shipped image and belongs in its own ADR and its own PR rather than riding inside a styling change. Recorded in ADR-004 so it is not lost.

### Also verified while in here

`npm audit --omit=dev` reports 2 moderate advisories against `react-router`. CI audits at `--audit-level=high`, so they do not fail the build, and they are pre-existing on `dev` and unrelated to this branch. Flagging rather than fixing, for the same scope reason.

PR #39, opened 2026-08-19
fix(frontend): make the layout fit the browser window at any size

NaimElijah:
## Problem

The interface did not adapt to the window. Maximised at 1536px, `/health-areas` left roughly a third
of the content area permanently empty; the "New Health Area" dialog nearly touched the top and bottom
of a short window with no way to scroll to what fell outside; and the seeded health areas rendered
`water_drop` and `fitness_center` as literal text, displacing each card's own title.

Three separate causes:

1. **Nine pages each set their own width cap** — four different values (`max-w-2xl` … `max-w-6xl`),
   nothing behind the choice. `/health-areas` was capped at 896px.
2. **`<main>` was safe from overflow only by accident.** It carried `overflow-auto`, which produces no
   scrollbar and reads as inert — but a flex item whose overflow is not `visible` already has an
   automatic minimum size of zero, so that class was the only thing keeping the shell shrinkable.
3. **Nothing constrained content coming from the database.** The `icon` field is free-form, the form
   labels it "Icon (emoji)", the seed data holds Material Symbols ligature names, and no icon font is
   loaded anywhere.

## Approach

Recorded in [ADR-005](docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md).

- **One container decides page width.** `PageContainer` offers 1280px for lists and grids, 768px for
  reading and form pages. The cap is on the measure, not the layout — with the sidebar and the shell's
  padding it does not bite until about a 1570px viewport, so below that a page fills the space.
- **The shell is shrinkable by construction, and says so.** `min-w-0` on `<main>` replaces the
  accidental `overflow-auto`. Content that then meets pressure resolves it itself: `truncate` on the
  navbar's user name, `break-words` on names and descriptions, a fixed clipping box around an icon,
  `flex-wrap` on header and button rows.
- **A dialog is bounded by its overlay.** The gutter is the overlay's padding, which is also what
  `max-h-full` resolves against; a `shrink-0` header over an `overflow-y-auto` body keeps the title in
  place while the form scrolls. It also gained `role="dialog"`, `aria-modal` and an `aria-labelledby`
  name from `useId`, since three modals share the health areas page.
- **Stored icons are reconciled on render, then corrected in the data.** `areaIconGlyph` falls back to
  the default for anything that is not a glyph, so the layout holds whatever is stored; `V5` then
  rewrites the six seeded rows as emoji, guarded on the old value so a user's own icon survives.

## Trade-offs accepted

- **A future unshrinkable child will now overflow visibly** instead of scrolling inside `<main>`. That
  is the intended failure — a visible bug beats a silent one — and the fix belongs on that child's own
  wrapper as an opt-in `overflow-x-auto`.
- **Three widths change deliberately:** daily check-in 672 → 768; notifications and progress history
  768 → 1280. Both lists are rows with right-aligned metadata that read better wide.
- **`aria-modal` overstates what the dialog does.** Nothing traps the keyboard, so Tab still walks out
  behind the scrim. Logged as an open question in `requirements.md` §6 rather than half-solved; the
  real answer is a native `<dialog>`, and ADR-005 records why it is not here.
- **Container queries were turned down.** Tailwind's breakpoints measure the viewport, but the content
  area is 240px narrower whenever the sidebar shows. ADR-005 names the threshold that would earn the
  dependency.

## Verified

`npm run lint`, `npm run check:colours`, `npm test` (54 pass, 19 new), `npm run build` — all green.
`mvn test` — 144 pass. `generate.py --check` — diagrams up to date.

jsdom has no layout engine and vitest runs with `css: false`, so no test here can see a width
(ADR-004). Checked in a real browser instead:

- `documentElement.scrollWidth === clientWidth` with zero offending elements at 486 / 684 / 1040 /
  1540px, across five pages, in both themes.
- At 1540px the content is 1252px of a 1300px area — the empty band is gone.
- Dialog in a 381px-tall window: 24px gutters, the body scrolls its 107px of overflow, the header does
  not move with it, Create is reachable.
- A 60-character unbroken name and `self_improvement` injected into a live card: the icon box stays
  32px and clips, the title wraps inside the card, still no overflow.
- `V5` dry-run against a running database inside a rolled-back transaction: six statements, one row
  each, all six emoji intact.

### DIFF
--- a/backend/src/main/resources/db/migration/V5__seed_health_area_icons_as_emoji.sql
+++ b/backend/src/main/resources/db/migration/V5__seed_health_area_icons_as_emoji.sql
@@ -1,0 +1,18 @@
+-- The seeded health areas carried Material Symbols ligature names in `icon` ('water_drop',
+-- 'fitness_center', ...), but no icon font is loaded anywhere in the frontend and the form labels the
+-- field "Icon (emoji)". The page drew those names as their own literal text. They are rewritten here as
+-- the emoji they were always meant to stand for.
+--
+-- A new migration rather than an edit to V2: Flyway stores a checksum per applied version, so changing
+-- an already-applied file makes every existing database refuse to start with a checksum mismatch.
+--
+-- Each row is matched on its id *and* on the value being replaced, so an account that has already set
+-- its own icon keeps it. Every glyph is a single code point with default emoji presentation, so nothing
+-- depends on how a variation selector survives the driver or the column.
+
+UPDATE health_areas SET icon = '💧' WHERE id = 'b0000000-0000-0000-0000-000000000001' AND icon = 'water_drop';
+UPDATE health_areas SET icon = '🏃' WHERE id = 'b0000000-0000-0000-0000-000000000002' AND icon = 'fitness_center';
+UPDATE health_areas SET icon = '🌙' WHERE id = 'b0000000-0000-0000-0000-000000000003' AND icon = 'bedtime';
+UPDATE health_areas SET icon = '🥗' WHERE id = 'b0000000-0000-0000-0000-000000000004' AND icon = 'restaurant';
+UPDATE health_areas SET icon = '🧘' WHERE id = 'b0000000-0000-0000-0000-000000000005' AND icon = 'self_improvement';
+UPDATE health_areas SET icon = '🌿' WHERE id = 'b0000000-0000-0000-0000-000000000006' AND icon = 'eco';
--- a/docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md
+++ b/docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md
@@ -1,0 +1,121 @@
+# ADR-005: One page width, and a shell that cannot be pushed wider than the window
+
+- **Status:** Accepted
+- **Date:** 2026-08-20
+- **Scope:** `frontend/` layout, plus one data-only backend migration. No new dependency, no runtime
+  behaviour outside the browser.
+
+## Context
+
+The SPA had no layout contract. Each page picked its own width cap as it was written, and by the time
+there were nine of them the same window showed the dashboard at 1152px, health areas at 896px and the
+daily check-in at 672px, with nothing a reader could reconstruct behind the choice. On a maximised
+1536px window `/health-areas` left roughly a third of the content area permanently empty.
+
+Three further things were true and none of them were written down:
+
+- **`<main>` was safe from overflow by accident.** It carried `overflow-auto`, which produces no
+  scrollbar (its height is auto, so it never overflows vertically) and looked inert. It was not: CSS
+  Flexbox §4.5 gives a flex item an automatic minimum size of zero when its overflow is anything other
+  than `visible`. Delete that class and the shell would start being dragged wider than the viewport by
+  its own content, with no clue in the diff as to why.
+- **Nothing constrained content that came from the database.** The six seeded health areas store
+  Material Symbols ligature names in `icon` — `water_drop`, `fitness_center`, `self_improvement` — but
+  no icon font is loaded anywhere and the form asks for an emoji. The card rendered the name verbatim
+  at `text-2xl`, where it displaced the area's own title and was clipped at the card's edge by `Card`'s
+  `overflow-hidden`. Storage, form and render path disagreed about what the field held.
+- **A dialog could not fit a short window.** `Modal` centred a panel with no maximum height and no
+  scroll region of its own, and guarded only its left and right edges with `mx-4`. A four-field form in
+  a 500px-tall viewport overflowed above and below the fold, and the part above it was unreachable
+  because nothing scrolled.
+
+None of this can be caught by a test here. jsdom has no layout engine and `vitest.config.ts` sets
+`css: false`, so every width, overflow and clipping question is invisible to the suite —
+[ADR-004](ADR-004-frontend-test-harness.md) records why a real browser in CI was deferred, and
+`requirements.md` §6 has carried that gap as an open question since.
+
+## Decision
+
+**One container decides page width.** `frontend/src/components/ui/PageContainer.tsx` offers two: the
+default 1280px (`max-w-7xl`) for lists, grids and dashboards, and 768px (`max-w-3xl`) for reading and
+form pages. All nine pages use it. The cap is on the measure — how long a line of text may get — not on
+the layout, so below roughly a 1570px viewport a default page simply fills the space the shell gives
+it. Horizontal padding stays on `<main>` so a page can never double it.
+
+**The shell is shrinkable by construction, and says so.** `min-w-0` on `<main>` replaces the accidental
+`overflow-auto`. A flex item's `min-width` defaults to `auto`, which resolves to its min-content width
+and silently overrules `flex-shrink`; `min-w-0` removes that floor. Content that then meets pressure
+resolves it itself — `truncate` on the navbar's user name, `break-words` on names and descriptions, a
+fixed `h-8 w-8 overflow-hidden` box around a health area's icon, `flex-wrap` on header rows and button
+rows.
+
+**A dialog is bounded by its overlay and scrolls its own body.** The overlay carries the gutter as
+padding; `max-h-full` resolves against that padded box, so the panel is never taller than the window;
+`flex flex-col` with a `shrink-0` header and an `overflow-y-auto` body keeps the title and close button
+in place while the form scrolls. It also gained `role="dialog"`, `aria-modal` and an `aria-labelledby`
+name derived with `useId`, since three modals share the health areas page.
+
+**Stored icons are reconciled on the render side, then corrected in the data.** `areaIcon.ts` falls back
+to the default glyph for anything starting with an ASCII word character, so the layout holds whatever
+the database contains; `V5__seed_health_area_icons_as_emoji.sql` then rewrites the six seeded rows as
+emoji. The render fix ships first, so at no point does the page depend on the migration having run.
+
+## Consequences
+
+**Easier.** Changing how wide pages are is one edit in one file. A new page inherits the decision
+instead of re-making it. A future overflow has one obvious place to look, and the reason `min-w-0` is
+there is written next to it rather than implied by a class that appears to do nothing.
+
+**Harder.** A genuinely unshrinkable future child — a wide data table — will now overflow visibly and
+give the document a horizontal scrollbar, where `overflow-auto` used to hide it inside `<main>`. That is
+the intended trade: a visible bug beats a silent one, and the fix belongs on that child's own wrapper as
+an opt-in `overflow-x-auto`.
+
+**Unchanged, and still unverified.** Nothing here is enforced by a test, and this ADR does not add a
+gate. Widths, wrapping, the dialog's scroll region and the absence of horizontal overflow were checked
+by hand in a browser at the widths listed in the pull request. The one mechanical assertion is
+`PageContainer.test.tsx`, which pins which cap each variant selects and explicitly claims nothing about
+the resulting layout.
+
+**Two widths changed deliberately.** The daily check-in goes from 672px to 768px and the notification
+and progress lists from 768px to 1280px. One reading cap and one default cap beat four arbitrary ones,
+and both list pages are rows with right-aligned metadata that read better wide.
+
+**`aria-modal` overstates what the dialog does.** It tells assistive technology the rest of the page is
+inert, but nothing traps the keyboard, so Tab still walks out behind the scrim, and focus is neither
+moved in on open nor restored on close. Recorded as an open question rather than half-solved.
+
+## Alternatives considered
+
+- **Leave the per-page caps and fix only the modal.** Rejected: the empty band on a wide window was the
+  reported symptom, and nine inconsistent caps is the kind of thing that gets copied into the tenth
+  page. *Revisit:* never — the inconsistency is the defect.
+- **No cap at all; content always fills the window.** Rejected: a 1920px window would give a
+  seventy-word line on the upgrade details page. The cap exists for the measure, not for the layout.
+- **`overflow-x: hidden` on the shell.** Rejected: it hides the symptom, and it clips any sticky or
+  popover child added later — the notification dropdown escapes its card today. *Revisit:* never.
+- **Container queries (`@tailwindcss/container-queries`).** Tailwind's breakpoints measure the viewport,
+  but the content area is 240px narrower whenever the sidebar is showing, so the column count changes
+  later than the numbers suggest — and at exactly `md` the sidebar appears and the content column drops
+  from 735px to 480px in one step. Container queries are the correct answer and cost a dependency.
+  *Revisit when* a second layout hits the same cliff, or the sidebar becomes collapsible and the
+  viewport-to-content offset stops being a constant.
+- **A native `<dialog>` with `showModal()`.** The correct answer to the focus trap: a real trap, real
+  inertness, native Escape and the top layer, with no hand-rolled key handling. Not adopted here because
+  the backdrop moves from a token-styled `<div>` to `::backdrop`, which lives in `index.css` — and
+  `check:colours` scans only `index.html` and `src/**/*.ts(x)`, so a colour used there would sit outside
+  the one guard that exists to have none. *Revisit when* the dialog needs a genuine keyboard workflow,
+  and extend `check:colours` to the stylesheet in the same change.
+- **Canonicalising a stored icon to exactly one grapheme with `Intl.Segmenter`.** Rejected: it does not
+  type-check under this project's `lib` setting, and its pre-Firefox-125 fallback splits by code point,
+  which strips a variation selector from `❤️` and a skin tone from `👍🏽`. The fixed-size clipping box
+  already makes the layout immune, so the render path only has to choose between the stored value and
+  the default.
+- **A mobile navigation drawer.** Out of scope. The sidebar is hidden below `md` deliberately and says
+  so; replacing it is a component, focus management and its own requirement.
+
+## Note
+
+Emoji are still written as literals in components — the sidebar's nav icons, the dashboard's stat
+glyphs, `EmptyState`. Most are decorative and lack `aria-hidden`, so a screen reader announces them.
+`Modal` now marks its own; the rest is a follow-up, not part of this decision.
--- a/docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md
+++ b/docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md
@@ -52,13 +52,18 @@ rows.
 **A dialog is bounded by its overlay and scrolls its own body.** The overlay carries the gutter as
 padding; `max-h-full` resolves against that padded box, so the panel is never taller than the window;
 `flex flex-col` with a `shrink-0` header and an `overflow-y-auto` body keeps the title and close button
-in place while the form scrolls. It also gained `role="dialog"`, `aria-modal` and an `aria-labelledby`
-name derived with `useId`, since three modals share the health areas page.
+in place while the form scrolls. It also gained `role="dialog"` and an `aria-labelledby` name derived
+with `useId`, since three modals share the health areas page — but deliberately **not** `aria-modal`,
+which is a claim about the rest of the page being inert that nothing here makes true.
 
 **Stored icons are reconciled on the render side, then corrected in the data.** `areaIcon.ts` falls back
-to the default glyph for anything starting with an ASCII word character, so the layout holds whatever
-the database contains; `V5__seed_health_area_icons_as_emoji.sql` then rewrites the six seeded rows as
-emoji. The render fix ships first, so at no point does the page depend on the migration having run.
+to the default for a value that is recognisably an icon *key* — an ASCII identifier, anchored at both
+ends so a keycap emoji such as `1️⃣` is not caught by its leading digit. It does not attempt to decide
+emoji-ness in general; it does not have to, because the icon is drawn in a fixed clipping box that no
+value can widen. The edit form uses the same predicate, so an undrawable value arrives as empty rather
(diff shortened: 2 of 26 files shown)


frontend/src/pages/DashboardPage.tsx
```diff
@@ -41,16 +42,16 @@ const DashboardPage: React.FC = () => {
   const streakEntries = Object.entries(data?.streaks ?? {});
 
   return (
-    <div className="max-w-6xl mx-auto space-y-6">
-      <div className="flex items-center justify-between">
-        <div>
+    <PageContainer className="space-y-6">
+      <div className="flex flex-wrap items-center justify-between gap-4">
+        <div className="min-w-0">
           <h1 className="text-2xl font-bold text-fg">
```
NaimElijah: **A long unbroken user name here now scrolls the whole document sideways.**

`user.name` is `@NotBlank` with no `@Size` (`AuthController.RegisterRequest`) and the column is `VARCHAR(255)`, so `name.split(' ')[0]` can be a 255-character unbreakable word. This `<h1>` is the one user-supplied string in the shell that has neither `truncate` nor `break-words` **and** is not inside a `Card` (whose `overflow-hidden` clips the equivalents on `/health-areas` and `/upgrades/:id`).

Before this PR `<main>` carried `overflow-auto`, so that overflow was contained as a scroll region inside `<main>`. With `min-w-0` replacing it the h1 overflows `<main>`, then the flex row, then the `min-h-screen` root, and the document gets a horizontal scrollbar. ADR-005 books this as a *future* risk ("a genuinely unshrinkable future child will now overflow visibly"), but it already exists today, and NFR-20 -- added in this same PR -- asserts the opposite: "does not scroll horizontally at any window width, **whatever a user has stored in it**".

The sibling cases got the fix; this one was missed.

```suggestion
          <h1 className="text-2xl font-bold text-fg break-words">
```

frontend/src/components/ui/areaIcon.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * An icon *key* starts with an ASCII word character; a glyph does not.
+ *
+ * The seeded areas were written against Material Symbols ligature names — `water_drop`,
+ * `fitness_center`, `self_improvement` — but no icon font is loaded anywhere, so the page drew those
+ * names as their own literal text. The form has always asked for an emoji. This is where the two
+ * readings are reconciled, on the render side, without trusting what is stored.
+ */
+const LOOKS_LIKE_AN_ICON_KEY = /^[\w-]/;
```
NaimElijah: **`/^[\w-]/` also rejects keycap emoji, which are exactly what the form asks for.**

`\w` is `[A-Za-z0-9_]`, so any icon whose first code point is an ASCII **digit** is treated as an icon key. The keycap emoji 0..9 (`U+0030..U+0039 U+FE0F U+20E3`) all start with an ASCII digit, so a user who types a keycap into the "Icon (emoji)" field silently gets the default target back -- the field accepts it, the API stores it, and the card refuses to draw it. (`#` and `*` keycaps survive only because those two characters are not `\w`.)

This also makes FR-41 stronger than the code: the requirement says "a health area whose stored icon **is not a glyph**", but the predicate actually implemented is "starts with an ASCII word character" -- which both over-rejects (keycaps) and under-rejects (`!`, `?`, trademark, middot are returned verbatim). `areaIcon.test.ts` covers neither direction.

Cheapest correction that keeps the same shape: anchor on the thing you are actually excluding -- a value that is *entirely* ASCII word characters/hyphens (`/^[\w-]+$/`), which matches every Material Symbols ligature name and no emoji at all.

frontend/src/components/layout/Navbar.tsx
```diff
@@ -10,20 +10,20 @@ const Navbar: React.FC = () => {
   const { user, logout } = useAuth();
 
   return (
-    <nav className="bg-surface border-b border-line h-16 flex items-center px-6 justify-between sticky top-0 z-30">
-      <Link to="/dashboard" className="flex items-center gap-2">
+    <nav className="bg-surface border-b border-line h-16 flex items-center gap-4 px-4 sm:px-6 justify-between sticky top-0 z-30">
```
NaimElijah, line 13: **Every item in this row is now `shrink-0`, so the navbar is the shell's own hard floor -- and it is wider than a phone.**

At `<sm` the user name is `hidden`, leaving: logo link (`shrink-0`, ~127px min-content) + `gap-4` (16) + ThemeToggle (`shrink-0`, 3x32 + gaps + `p-0.5` ~ 104) + `gap-2` (8) + bell (`shrink-0`, `w-9` = 36) + `gap-2` (8) + Logout (`shrink-0`, ~70) + `px-4` (32) ~ **401px**. Below that the nav overflows the viewport and, since it is a block child of the `min-h-screen` root, the document scrolls horizontally -- on a 375px iPhone and a 360px Android, both above the 320px floor.

The PR's own verification starts at 486px, so this width band was never exercised. It does not contradict `min-w-0` on `<main>` (the nav is not in `<main>`), but it does contradict NFR-20 as written -- "at any window width".

Either narrow the claim in `requirements.md` to the range actually verified, or let something here shrink: dropping `shrink-0` from the logo link and giving it `min-w-0 truncate` costs nothing and buys ~95px.

frontend/src/pages/DailyCheckinPage.tsx
```diff
@@ -82,7 +83,7 @@ const DailyCheckinPage: React.FC = () => {
   }
 
   return (
-    <div className="max-w-2xl mx-auto">
+    <PageContainer width="narrow">
```
NaimElijah, line 86: **A tenth width cap survives, in this file, at line 76.**

The `submitted` success state returns before this line and keeps its own cap:

```tsx
if (submitted) {
  return (
    <div className="max-w-xl mx-auto text-center py-20">
```

`max-w-xl` is 576px -- a fifth value that `PageContainer` does not offer, sitting in the one page that was already being converted. ADR-005 states "All nine pages use it" and warns that "nine inconsistent caps is the kind of thing that gets copied into the tenth page"; both the ADR and FR-38 are currently untrue of this branch.

Wrap it in `<PageContainer width="narrow">` like the main return, or if the narrower measure is deliberate for a centred confirmation, that is a third variant and belongs in `widthClasses` rather than inline.

frontend/src/pages/HealthAreasPage.tsx
```diff
@@ -77,28 +79,31 @@ const HealthAreasPage: React.FC = () => {
...
-                    {area.upgradeCount !== undefined && (
-                      <p className="text-xs text-fg-subtle">{area.upgradeCount} upgrade{area.upgradeCount !== 1 ? 's' : ''}</p>
-                    )}
-                  </div>
+              <div className="flex min-w-0 items-start gap-2">
+                <span
+                  className="flex h-8 w-8 shrink-0 items-center justify-center overflow-hidden text-2xl leading-none"
+                  aria-hidden="true"
+                >
+                  {areaIconGlyph(area.icon)}
```
NaimElijah, line 107: **The card resolves the icon but the edit form does not, so what you see and what you edit disagree.**

`openEdit` (line 66) and both `Input`s (lines 119, 132) still bind the **raw** stored value:

```tsx
setForm({ ..., icon: area.icon ?? '', ... });
```

On any database where `V5` has not run -- or for any area a user saved with a non-glyph value -- the card renders the default glyph while the edit dialog shows `water_drop`, and pressing "Save Changes" writes `water_drop` straight back. The render fix and the data fix between them never close the write path.

This is the altitude question behind the whole icon change: the field is described in the PR as "free-form, validated nowhere", but the fix lands entirely in the view. The boundary is `HealthAreaRequest.icon` -- a constraint there (or `areaIconGlyph` applied in `openEdit`) is what stops the next write from re-creating the bug the migration just cleaned up.

frontend/src/components/ui/PageContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * So the cap is on the measure, not on the layout: `wide` is roughly a hundred characters at the body
+ * size, `narrow` roughly sixty. Given the 240px sidebar and the shell's padding, `wide` does not begin
+ * to bite until about a 1570px viewport — below that a page simply fills the space.
+ */
+const widthClasses: Record<NonNullable<PageContainerProps['width']>, string> = {
+  wide: 'max-w-7xl',
+  narrow: 'max-w-3xl',
+};
+
+const PageContainer: React.FC<PageContainerProps> = ({ children, width = 'wide', className = '' }) => (
```
NaimElijah, line 35: **The component's doc comment is attached to `widthClasses`, so the exported component has none.**

The block at lines 15-26 -- "The one place a page's width is decided", the history of the four caps, the 1570px threshold -- sits immediately above `const widthClasses`, which is a private lookup table. `PageContainer`, the default export and the only thing any other file touches, is undocumented: hovering it in an editor shows nothing, and hovering the private map shows the page-width essay.

`~/.claude/CLAUDE.md`: "Write a doc comment on every public or exported type, class, function, and module: its purpose, its contract ... Private helpers get one only when the reason they exist is not self-evident."

Move the block below `widthClasses` and onto `const PageContainer`.

frontend/src/components/ui/Modal.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    it('GivenAnOpenModal_WhenTheBackdropIsClicked_ThenItCloses', () => {
+      const onClose = vi.fn();
+      const { container } = renderModal(onClose);
+
+      fireEvent.click(container.firstElementChild!.firstElementChild!);
+
+      expect(onClose).toHaveBeenCalledTimes(1);
+    });
+
+    it('GivenAModalThatHasClosed_WhenEscapeIsPressed_ThenNothingListensForIt', () => {
```
NaimElijah: **This test cannot fail, so it does not pin the behaviour its name claims.**

`Openable` holds `open` in state and `onClose` is `() => setOpen(false)`. If the listener were *still bound* after close, the second `keyDown` would call `setOpen(false)` on an already-false state -- and the assertion `queryByRole('dialog')` is `null` either way. Delete the `if (isOpen)` guard from `Modal`'s effect and this test still passes; delete the cleanup return and it still passes. There is no input that distinguishes the two worlds.

To actually observe it, count the calls on a spy across the transition:

```tsx
const onClose = vi.fn();
const Openable = () => {
  const [open, setOpen] = useState(true);
  return <Modal isOpen={open} onClose={() => { onClose(); setOpen(false); }} title="...">
    <p>body</p>
  </Modal>;
};
// after two keyDowns:
expect(onClose).toHaveBeenCalledTimes(1);
```

Separately, the doc comment on `Openable` (line 25) says "Opens a modal from a button" -- there is no button; it starts open and only closes.

frontend/src/components/ui/Modal.tsx
```diff
@@ -32,20 +50,27 @@ const Modal: React.FC<ModalProps> = ({ isOpen, onClose, title, children }) => {
...
   return (
-    <div className="fixed inset-0 z-50 flex items-center justify-center">
+    <div className="fixed inset-0 z-50 flex items-center justify-center p-4 sm:p-6">
       <div className="absolute inset-0 bg-overlay/50" onClick={onClose} />
-      <div className="relative z-10 w-full max-w-lg rounded-xl bg-surface border border-line-strong shadow-xl mx-4">
-        <div className="flex items-center justify-between px-6 py-4 border-b border-line">
-          <h2 className="text-lg font-semibold text-fg-muted">{title}</h2>
+      <div
+        role="dialog"
+        aria-modal="true"
```
NaimElijah: **`aria-modal="true"` without inertness is a net regression for screen-reader users, not a neutral half-measure.**

The gap is honestly recorded in the docstring, ADR-005 and requirements section 6 -- but the accounting understates it. Before this PR the panel had no `role`, so assistive technology treated it as ordinary page content: reachable, announced, consistent with where Tab actually went. `aria-modal="true"` now tells AT that everything outside this node is inert and drops it from the virtual buffer, while Tab still walks the focus out there. A screen-reader user who tabs past the last field lands on controls their reader refuses to announce -- silence, with focus somewhere invisible.

`role="dialog"` + `aria-labelledby` are pure wins and should stay. `aria-modal` is the one attribute whose promise is unbacked; until focus is trapped (or `<dialog>.showModal()` lands), dropping just that attribute leaves the dialog correctly announced and honest about its behaviour.

docs/ADRs/ADR-005-one-page-width-and-a-shell-that-cannot-overflow.md
```diff
@@ -0,0 +1,121 @@
...
+the intended trade: a visible bug beats a silent one, and the fix belongs on that child's own wrapper as
+an opt-in `overflow-x-auto`.
+
+**Unchanged, and still unverified.** Nothing here is enforced by a test, and this ADR does not add a
+gate. Widths, wrapping, the dialog's scroll region and the absence of horizontal overflow were checked
+by hand in a browser at the widths listed in the pull request. The one mechanical assertion is
+`PageContainer.test.tsx`, which pins which cap each variant selects and explicitly claims nothing about
+the resulting layout.
+
+**Two widths changed deliberately.** The daily check-in goes from 672px to 768px and the notification
```
NaimElijah: **Eight of the nine pages changed width, not two -- and four of them were widened without a word.**

Unlisted here and in the PR description:

| page | before | after |
|---|---|---|
| `ActiveUpgradesPage` | `max-w-4xl` 896 | 1280 |
| `PlannedUpgradesPage` | `max-w-4xl` 896 | 1280 |
| `UpgradeBacklogPage` | `max-w-4xl` 896 | 1280 |
| `DashboardPage` | `max-w-6xl` 1152 | 1280 |

That matters beyond bookkeeping, because all three upgrade pages keep `grid sm:grid-cols-2` unchanged. At the 1568px viewport where the cap starts to bite, the extra 384px is split between **two** cards, giving ~630px-wide `UpgradeCard`s whose titles are `truncate`d to one line over a 2-line clamped description -- the same "roughly a third of the content area permanently empty" symptom, relocated from the page margin into the cards. `/health-areas` got `xl:grid-cols-3` for exactly this reason; the upgrade grids did not.

Either give them an `xl:grid-cols-3` too, or record the widening and the reason it is fine -- an ADR that documents the present has to name all of what changed.

frontend/src/components/ui/PageContainer.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * The one place a page's width is decided.
+ *
+ * Pages used to cap themselves, with four different values across nine pages and nothing behind the
+ * choice — which is how `/health-areas` came to sit 896px wide in the middle of a 1536px window. A page
+ * now fills whatever the shell gives it and stops only where a line of text stops being comfortable to
+ * read.
+ *
+ * So the cap is on the measure, not on the layout: `wide` is roughly a hundred characters at the body
```
NaimElijah: **The character counts are off by ~1.6x, and they are the entire justification for the two caps.**

At the 16px body size an average proportional character is ~8px wide (Tailwind's own `max-w-prose` is `65ch`, about 520px). So:

- `wide` = `max-w-7xl` = 1280px, about **160** characters, not "roughly a hundred"
- `narrow` = `max-w-3xl` = 768px, about **96** characters, not "roughly sixty"

"Roughly sixty characters" would be about 480px. Since the stated principle is "the cap is on the measure, not on the layout", the numbers are the argument -- and as written they claim a typographic measure the values do not deliver. Either correct the figures, or if ~60ch is genuinely the target for reading pages, `narrow` is the value that should move.

backend/src/main/resources/db/migration/V5__seed_health_area_icons_as_emoji.sql
```diff
@@ -0,0 +1,18 @@
...
+-- field "Icon (emoji)". The page drew those names as their own literal text. They are rewritten here as
+-- the emoji they were always meant to stand for.
+--
+-- A new migration rather than an edit to V2: Flyway stores a checksum per applied version, so changing
+-- an already-applied file makes every existing database refuse to start with a checksum mismatch.
+--
+-- Each row is matched on its id *and* on the value being replaced, so an account that has already set
+-- its own icon keeps it. Every glyph is a single code point with default emoji presentation, so nothing
+-- depends on how a variation selector survives the driver or the column.
+
```
NaimElijah, line 12: **The migration and `areaIcon.ts` redefine what `icon` holds; the two Java records still document the old meaning.**

Untouched by this PR, in the same field's own contract:

- `HealthAreaDto.java:14` -- `@param icon icon key the frontend resolves to a glyph, may be null`
- `HealthAreaRequest.java:14` -- `@param icon icon key, optional`

After this change the frontend does the exact opposite: `areaIconGlyph` *rejects* anything that looks like a key and substitutes the default. The wire contract now says "icon key" while the only consumer treats an icon key as invalid input -- the next person to seed or import data has a documented reason to write `water_drop` again.

`~/.claude/CLAUDE.md`: "Documentation changes in the same commit as the behaviour it describes." Both javadocs should read something like "emoji glyph; anything else is replaced by the default on render".

(The SQL itself checks out: valid UTF-8, no BOM, ids match `V2`, and the `AND icon = '...'` guard correctly spares a user's own value.)

frontend/src/pages/HealthAreasPage.tsx
```diff
@@ -77,28 +79,31 @@ const HealthAreasPage: React.FC = () => {
...
-              <div className="flex items-start justify-between">
-                <div className="flex items-center gap-2">
-                  <span className="text-2xl">{area.icon ?? '🎯'}</span>
-                  <div>
-                    <h3 className="font-semibold text-fg">{area.name}</h3>
-                    {area.upgradeCount !== undefined && (
-                      <p className="text-xs text-fg-subtle">{area.upgradeCount} upgrade{area.upgradeCount !== 1 ? 's' : ''}</p>
-                    )}
-                  </div>
+              <div className="flex min-w-0 items-start gap-2">
```
NaimElijah: **`min-w-0` here is a no-op.**

This `<div>` is a plain block child of `Card`'s inner `p-6` wrapper, not a flex item, so its `min-width` was already `0`; the automatic-minimum-size rule that makes `min-w-0` load-bearing on `<main>` and on the inner `<div className="min-w-0">` at line 98 does not apply to it.

In a PR whose thesis is that `min-w-0` is load-bearing and should say so (see the `Layout` docstring and the new architecture section), a decorative copy of it in a place where it does nothing is the one thing that undermines the convention -- a reader can no longer tell which occurrences carry weight. Drop it from line 91; the one on line 98 is the real one.

frontend/src/pages/UpgradeDetailsPage.tsx
```diff
@@ -122,7 +123,7 @@ const UpgradeDetailsPage: React.FC = () => {
   if (error || !upgrade) return <p className="text-danger-fg text-center py-10">Failed to load this upgrade.</p>;
 
   return (
-    <div className="max-w-3xl mx-auto space-y-6">
+    <PageContainer width="narrow" className="space-y-6">
```
NaimElijah, line 126: **The upgrade title on this page is clipped rather than wrapped -- `break-words` was applied to health-area names but not here.**

`upgrade.title` is `VARCHAR(255)` user text. Line 137 renders it as `<h1 className="text-2xl font-bold text-fg">` with no `break-words`, and `motivation` (141) and `successCriteria` (146) are the same. All three sit inside `Card`, whose `overflow-hidden` means the failure is not a scrollbar -- it is silent truncation: a long unbroken title simply disappears past the card's right edge with no ellipsis and no way to read it.

This is the same class of gap the PR closed on `/health-areas` ("`break-words` on names and descriptions"); the details page holds the same data and did not get it. `break-words` on the `h1` and the two paragraphs matches the treatment already chosen elsewhere.

NaimElijah:
## Review addressed — 5 commits

All 13 findings acted on. Three of them were things I had asserted and not checked, so they are worth
naming rather than burying:

**The navbar had a 436px floor.** Making every control `shrink-0` put a hard minimum under the whole
document, so NFR-20's "at any window width" was simply false below that. My browser sweep started at
486px because Chrome will not size a window narrower — the band was unreachable by resizing and needed
device emulation. The brand now gives instead (`min-w-0` + `truncate`); measured at a real 320×568 and
360×740: no overflow, nothing past the viewport edge, logout still reachable.

**`aria-modal` was a net regression, not a neutral half-measure.** It removes the page behind from a
screen reader's buffer while focus can still tab there. Withheld until a focus trap exists; `role` and
the accessible name stay, and a test pins the omission as a decision.

**One test could not fail.** It asserted the dialog was still absent after Escape on a closed modal —
true whether or not the listener is bound. Replaced with an assertion on the callback, and verified by
removing the `if (isOpen)` guard: old test passed, new one fails.

### The rest

| Finding | Done |
|---|---|
| Dashboard greeting could scroll the document | `break-words` on the `h1` |
| `/^[\w-]/` misclassifies keycap emoji | Anchored full-token match; `1️⃣`, `👍🏽`, `❤️` covered by tests |
| Edit form re-persists the undrawable icon | `openEdit` uses the same predicate; arrives empty |
| Check-in success panel bypasses `PageContainer` | Converted — the ADR's "all nine pages" is true again |
| `PageContainer` doc on the private helper | Moved to the exported component |
| Four pages widened but unnamed in the ADR | Corrected: eight of nine changed, and it says which |
| Wide cards, `sm:grid-cols-2` kept | Third column at `xl` on the three upgrade lists |
| Character-count claim off by ~1.6× | 1280px ≈ 160 chars, 768px ≈ 95 — corrected |
| DTO javadoc still says "icon key" | Both `HealthAreaDto` and `HealthAreaRequest` rewritten |
| No-op `min-w-0` on a block child of `Card` | Removed |
| Upgrade title/motivation/criteria clipped | `break-words` on all four fields |

### Deliberately not changed

The predicate still returns `!` and `™` verbatim. It rejects what is recognisably an icon *key*, not
everything that is not an emoji — deciding emoji-ness properly needs grapheme segmentation, which costs
a `lib` change and mangles skin tones where it is unavailable, and buys nothing when the icon is drawn
in a fixed clipping box no value can widen. That boundary is now two tests instead of an accident.

### Verified

`lint`, `check:colours`, `test` (61 pass, up from 54), `build` — green. `mvn test` — 144 pass.
`generate.py --check` — up to date. Browser: no horizontal overflow at 320, 360, 486, 684, 1040 or
1540px; the dialog fits 320×568 with gutters on all four sides.

Issue #7, opened 2026-07-09
Upload Started Repo to GitHub


NaimElijah:
Upload Started Repo to GitHub.
After this, will continue enhancing and upgrading this project.

NaimElijah:
Done.

PR #42, opened 2026-08-23
fix(api): a framework exception gets its own status, not 500

### LINKED ISSUE #22: fix: an unbindable request body returns 500 instead of 400

NaimElijah:
Closes #22.

## Problem

A request body the API cannot bind returned **500**, not 400:

```
POST /api/upgrades  {"title":"x","type":"NOT_A_TYPE"}
→ 500 {"status":500,"message":"Internal server error","path":"/api/upgrades",...}
```

`GlobalExceptionHandler` is consulted by `ExceptionHandlerExceptionResolver`, which runs **before**
Spring's own `DefaultHandlerExceptionResolver`. So its catch-all `@ExceptionHandler(Exception.class)`
did not merely handle what was left over — it claimed every framework exception first, and the status
Spring defines for that exception never applied. Review of the first commit found that #22 was one
instance of that, not the whole of it. On `UpgradeController` alone, all verified:

| Request | Was | Now |
|---|---|---|
| `POST /api/upgrades` `{"type":"NOT_A_TYPE"}` | 500 | 400 |
| `GET /api/upgrades/not-a-uuid` | 500 | 400 |
| `GET /api/upgrades?status=BOGUS` | 500 | 400 |
| `PATCH /api/upgrades/{id}` | 500 | 405 |
| `POST /api/upgrades` `Content-Type: text/plain` | 500 | 415 |
| a route that matches nothing | 500 | 404 |

Every one is the caller's mistake reported as the server's failure, and every one was counted as an
unhandled failure — the noise a real 500 then has to be found in. `handleGeneral` also logged nothing
at all, so an unexpected 500 left no server-side trace beyond the access log.

## Approach

`GlobalExceptionHandler` extends `ResponseEntityExceptionHandler`, inheriting Spring's mapping for
every standard MVC exception, and overrides `handleExceptionInternal` to re-clothe the result in this
API's `ErrorResponse` body rather than a `ProblemDetail`. Three hooks are overridden where the
inherited message is not good enough (`handleMethodArgumentNotValid` keeps the field → message map,
`handleHttpMessageNotReadable` and `handleTypeMismatch` name the class of mistake); everything else
takes the status' own reason phrase, which cannot leak the exception's detail.

The catch-all stays for exceptions that are genuinely ours and logs at ERROR with the stack trace. A
4xx logs at DEBUG — a caller's mistake is not an incident.

The first commit added a single handler for `HttpMessageNotReadableException`, which is the approach
that produced the bug in the first place: "add the handler when you meet one" means a user meets a 500
first, and the list is the framework's, not ours. That commit is kept in the history rather than
squashed, because the review that rejected it is the reason the second one exists.

## Trade-off accepted

The advice's behaviour is no longer readable from the class alone — part of the contract lives in a
superclass whose handled types change with the Spring version. A Spring upgrade that adds a type, or
changes one's status, changes this API's wire contract; `GlobalExceptionHandlerTest` is what turns that
into a failing test rather than a silent change.

Six statuses changed for clients. No client branches on any of them (`frontend/src/api/client.ts`
inspects 401 and nothing else), so this was taken in one step rather than staged.

Recorded as [ADR-006](docs/ADRs/ADR-006-framework-exceptions-through-responseentityexceptionhandler.md),
with the rejected alternatives — a handler per exception, deleting the catch-all, adopting RFC 9457
`ProblemDetail` — and what would make each worth revisiting.

## Verification

Reproduced first, both times. The new tests against the unfixed handler:

```
malformedJsonBody_isBadRequest                Status expected:<400> but was:<500>
unknownEnumValueInBody_isBadRequest           Status expected:<400> but was:<500>
unexpectedFailure_isLoggedWithItsStackTrace   Expected size: 1 but was: 0
```

then the four remaining 500s in the table above, probed against the first commit before the second was
written. After the fix, `mvn test`:

```
Tests run: 155, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The exception contract went from 8 assertions to 18, including: bean-validation's field errors (pinned
for the first time), a client mistake logging nothing at ERROR, and the handler resolver being built
the way Spring builds it at startup — so an ambiguous mapping fails in `mvn test` instead of at boot,
where only a database-backed IT would have seen it.

## Documentation

- `docs/ADRs/ADR-006-…` — the decision, its consequences and the alternatives.
- `docs/requirements/requirements.md` — NFR-7 rewritten to what the code now honours; NFR-6 states the
  one exception to "personal data is never logged" (a 5xx stack trace) instead of being quietly
  contradicted by it.
- `backend/CLAUDE.md` — framework exceptions are not ours to map; override the hook, do not add an
  `@ExceptionHandler` for a type the parent already maps.

## Not in this PR

The review also found there is no correlation id anywhere in the backend, so a reported 500 cannot be
joined to its log line. That is a cross-cutting concern with its own design — filed as #43.


backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
```diff
@@ -101,10 +124,13 @@ public ResponseEntity<ErrorResponse> handleAccessDenied(AccessDeniedException ex
      * Catch-all for anything unmapped: 500 with a fixed message.
      *
      * <p>The exception's own message is deliberately withheld — an unanticipated failure can carry a
-     * stack detail, a SQL fragment or a value that must not reach a client.
+     * stack detail, a SQL fragment or a value that must not reach a client. It is logged instead, since
+     * withholding the cause is only safe if the server keeps it: a 500 nobody can diagnose is worse than
+     * the leak it avoids.
      */
     @ExceptionHandler(Exception.class)
```
NaimElijah, line 180: **The same bug is still live for every other framework exception — verified by running it.**

`HttpMessageNotReadableException` was one of several exceptions the catch-all steals from `DefaultHandlerExceptionResolver`. I ran a probe through this exact `MockMvc` rig against `UpgradeController` with the patched advice:

```
GET   /api/upgrades/not-a-uuid                -> 500   (Spring maps TypeMismatch -> 400)
GET   /api/upgrades?status=BOGUS              -> 500   (Spring maps TypeMismatch -> 400)
PATCH /api/upgrades/{id}                      -> 500   (Spring maps -> 405)
POST  /api/upgrades  Content-Type: text/plain -> 500   (Spring maps -> 415)
```

A caller who fat-fingers an id in the URL is still told the server broke. `backend/CLAUDE.md` now writes the whack-a-mole down as the process ("Add the handler when you meet one"), which is the wrong depth: one class of framework exception was patched and the mechanism that swallows the rest is untouched.

The deep fix is to stop the catch-all claiming what Spring already maps — either extend `ResponseEntityExceptionHandler` (it covers the whole standard set with the right statuses, and `handleExceptionInternal` is the single place to attach the body), or gate the catch-all:

```java
@ExceptionHandler(Exception.class)
public ResponseEntity<ErrorResponse> handleGeneral(Exception ex, HttpServletRequest req) {
    if (ex instanceof ErrorResponse er) {   // Spring 6 marker, carried by most framework exceptions
        return ResponseEntity.status(er.getStatusCode()).body(/* ... */);
    }
    // ... existing 500 path
}
```

Either way the trap closes once instead of once per exception discovered in production.

backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
```diff
@@ -101,10 +124,13 @@ public ResponseEntity<ErrorResponse> handleAccessDenied(AccessDeniedException ex
...
     public ResponseEntity<ErrorResponse> handleGeneral(Exception ex, HttpServletRequest req) {
+        log.error("Unhandled exception for {} {}", req.getMethod(), req.getRequestURI(), ex);
```
NaimElijah, line 182: **This turns routine client mistakes into ERROR-with-stack-trace — the noise the PR set out to remove.**

Because the catch-all still claims the framework exceptions above, every one of them now also logs at ERROR with a full stack. Captured from the probe run:

```
ERROR GlobalExceptionHandler -- Unhandled exception for GET /api/upgrades/not-a-uuid
ERROR GlobalExceptionHandler -- Unhandled exception for PATCH /api/upgrades/00000000-...
ERROR GlobalExceptionHandler -- Unhandled exception for POST /api/upgrades   (text/plain)
```

`~/.claude/CLAUDE.md`: *"Log levels mean something: ERROR is actionable now."* A bad UUID in a path is not actionable. The PR body argues the old behaviour was "noise that makes a genuine 500 harder to spot" — after this change the same requests still return 500 **and** now flood the ERROR channel with stack traces.

`/api/auth/**` is `permitAll()` in `SecurityConfig`, so this is reachable unauthenticated: `POST /api/auth/login` with `Content-Type: text/plain` → 500 + ERROR + stack, no token needed.

Honouring the framework's status (comment above) fixes this too, since those exceptions stop reaching `handleGeneral`.

NaimElijah, line 182: **Logging the whole exception contradicts NFR-6 and the global no-personal-data rule — and the new test asserts it.**

`docs/requirements/requirements.md` NFR-6: *"Personal data is never written to logs"*. `~/.claude/CLAUDE.md`: *"Never log credentials, tokens, or personal data. Redact at the logging boundary, not at each call site."*

`log.error(..., ex)` writes the message and the full cause chain. In this app that chain routinely carries user data: a Postgres `DataIntegrityViolationException` surfaces as `Key (email)=(alice@example.com) already exists`, and a Hibernate failure carries the SQL with bound parameters — upgrade titles like "Quit drinking" or "Therapy weekly", which `SpringDomainEventPublisher` goes out of its way *not* to log for exactly this reason (its comment: *"personal data in a health application"*).

The new test at `GlobalExceptionHandlerTest:175` makes the point involuntarily: it feeds `IllegalStateException("jdbc://user:hunter2@db/prod")` and asserts the throwable **is** kept. `unexpectedFailure_isInternalServerErrorAndSaysNothingElse` asserts `hunter2` never reaches the client; the new test asserts it reaches the log.

Two things are conflated: an unmapped exception needs a **stack trace** to be diagnosable; it does not need its **message** verbatim. Either redact at the logging boundary as the rule requires, or — since this is a deliberate trade-off against a stated NFR — record it as an ADR and amend NFR-6, rather than leaving the two documents contradicting each other.

backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
```diff
@@ -90,6 +99,20 @@ public ResponseEntity<ErrorResponse> handleValidation(MethodArgumentNotValidExce
...
+    /**
+     * Maps a request body that cannot be bound to 400 — malformed JSON, an unknown enum constant, a
+     * string where a number is expected.
+     *
+     * <p>The client's input was wrong, not the server, and the failure happens before any controller
+     * method is entered. The parser's own message is withheld: it names the target type, the offending
+     * value and the byte offset, which describes the API's internals rather than the caller's mistake.
+     */
+    @ExceptionHandler(HttpMessageNotReadableException.class)
+    public ResponseEntity<ErrorResponse> handleNotReadable(HttpMessageNotReadableException ex, HttpServletRequest req) {
```
NaimElijah: **`handleNotReadable` applies the opposite of the rule this PR just wrote for `handleGeneral`.**

The javadoc two methods down now says: *"withholding the cause is only safe if the server keeps it: a 500 nobody can diagnose is worse than the leak it avoids."* This handler withholds the parser's message from the client and then drops it entirely — nothing is logged at any level. Support has no way to answer "why is my POST a 400?".

That matters more than it looks, because `HttpMessageNotReadableException` is not only a client fault: Spring wraps an `IOException` raised while reading the body into the same exception (*"I/O error while reading input message"*). A client disconnect or a server-side read failure now returns a clean 400 with zero trace — a transport problem reported as the caller's mistake, leaving nothing behind.

`log.debug("Unreadable request body for {} {}", req.getMethod(), req.getRequestURI(), ex)` costs nothing and makes the handler consistent with its neighbour.

docs/requirements/requirements.md
```diff
@@ -138,7 +138,7 @@ nothing is shared between accounts.
 | NFR-4 | A token is valid for 24 hours and is not refreshable | `app.jwt.expiration` |
 | NFR-5 | The user behind a token is re-loaded on every request, so a deleted account stops working immediately | `JwtAuthenticationFilter` |
 | NFR-6 | Personal data is never written to logs; event publication logs the type and timestamp only | `SpringDomainEventPublisher` |
-| NFR-7 | Every failure maps to a defined HTTP status: 404 not found, 422 rule violation, 409 conflict, 400 invalid input, 403 denied | `GlobalExceptionHandler`, `GlobalExceptionHandlerTest` |
+| NFR-7 | Every failure maps to a defined HTTP status: 404 not found, 422 rule violation, 409 conflict, 400 invalid input (a failed constraint or a body that cannot be bound), 403 denied; a 500 carries no detail beyond the status and is logged server-side | `GlobalExceptionHandler`, `GlobalExceptionHandlerTest` |
```
NaimElijah: **NFR-7 as rewritten is not true, and the open question that tracked it was deleted rather than narrowed.**

The row claims *"Every failure maps to a defined HTTP status"* and names `GlobalExceptionHandler` / `GlobalExceptionHandlerTest` as the enforcers. Verified against the patched code: `GET /api/upgrades/{bad-uuid}` → 500, `PATCH /api/upgrades/{id}` → 500, `POST` with `text/plain` → 500. Those are failures with a defined status the API does not return.

Separately, the "400 invalid input (**a failed constraint** or a body that cannot be bound)" half names a test that does not exist: `grep -rn "MethodArgumentNotValid\|handleValidation\|Validation failed" src/test` returns nothing. `handleValidation` is pinned by no test anywhere, while this row and the class javadoc both assert the whole mapping is pinned.

`~/.claude/CLAUDE.md`: *"Where a requirement is enforced in code, name the test or class that enforces it, so the document can be checked rather than trusted."* As written this row cannot be checked.

The removed open question should be **narrowed** to the exceptions still unhandled, not deleted — issue #22 is only partly closed. And the MockMvc rig this PR introduces makes the missing bean-validation test a two-line addition.

backend/src/test/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandlerTest.java
```diff
@@ -90,6 +118,39 @@ void accessDenied_isForbidden() {
...
+    @Test
+    void malformedJsonBody_isBadRequest() throws Exception {
+        mockMvc.perform(post("/api/upgrades")
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .content("{\"title\": "))
+                .andExpect(status().isBadRequest());
+    }
+
+    @Test
+    void unbindableBody_saysNothingAboutTheParser() throws Exception {
```
NaimElijah: **This test passes just as happily against the unfixed code, so it pins nothing.**

It never asserts the status, and the strings it excludes — `NOT_A_TYPE`, `com.fasterxml`, `UpgradeType` — are equally absent from the old 500 body (`{"status":500,"message":"Internal server error",...}`). Reverting `handleNotReadable` entirely leaves this test green. It also re-sends the byte-identical request from `unknownEnumValueInBody_isBadRequest` three lines up.

Fold the two into one test that asserts what actually changed:

```java
mockMvc.perform(post("/api/upgrades")
                .contentType(MediaType.APPLICATION_JSON)
                .content("{\"title\":\"Walk daily\",\"type\":\"NOT_A_TYPE\"}"))
        .andExpect(status().isBadRequest())
        .andExpect(jsonPath("$.message").value("Malformed request body"))
        .andExpect(content().string(not(containsString("NOT_A_TYPE"))));
```

Nothing currently pins the 400 body's `message` or `path` either — the wire contract this PR adds is asserted only as "some 4xx".

backend/src/test/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandlerTest.java
```diff
@@ -1,37 +1,65 @@
...
     @BeforeEach
     void setUp() {
         handler = new GlobalExceptionHandler();
         MockHttpServletRequest mockRequest = new MockHttpServletRequest();
         mockRequest.setRequestURI("/api/upgrades/42");
         request = mockRequest;
+
+        // The service and mapper are never reached: binding fails before the handler method is called.
+        mockMvc = MockMvcBuilders
+                .standaloneSetup(new UpgradeController(mock(UpgradeService.class), mock(UpgradeWebMapper.class)))
```
NaimElijah: **The shared kernel's test is now wired to a bounded context's controller, and the rig is rebuilt for all nine tests.**

Two things:

1. `common`'s advice test now imports `upgrade.adapter.in.web.UpgradeController`, `UpgradeWebMapper` and `upgrade.application.UpgradeService`. `HexagonalArchitectureTest` runs with `ImportOption.DoNotIncludeTests`, so nothing catches it — but NFR-9's "contexts form an acyclic graph" is the property being informally broken, and adding a constructor parameter to `UpgradeController` now breaks a test in `common` for reasons unconnected to it. A four-line `@RestController` fixture declared inside this test class, taking a record with an enum field, exercises the identical resolver path and owns its own contract.

2. `@BeforeEach` builds the MockMvc rig plus two Mockito mocks before **every** test; six of the nine never touch `mockMvc`. A field initialiser, or a `@Nested` class grouping the MockMvc cases, avoids the repeat.

Worth knowing about the rig itself: `standaloneSetup` does not use the application's `ObjectMapper`. The body it actually produced in my run was `{"status":400,...,"timestamp":[2026,8,23,19,14,28,946981200],"fieldErrors":null}` — an array timestamp, not what production serialises. Whether an unknown enum constant throws is likewise decided by the default mapper here, not the one the app runs.

backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
```diff
@@ -90,6 +99,20 @@ public ResponseEntity<ErrorResponse> handleValidation(MethodArgumentNotValidExce
...
+        return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+                .body(new ErrorResponse(HttpStatus.BAD_REQUEST.value(), "Malformed request body", req.getRequestURI()));
```
NaimElijah: **Adjacent to this line: the `ErrorResponse` javadoc below is wrong, and this handler ships the proof.**

Line 141: *"`fieldErrors` is populated only for validation failures and is omitted from the JSON otherwise (it is null, and null fields are not serialized)."*

There is no `@JsonInclude(NON_NULL)` on the class and no Jackson configuration anywhere — `grep -rn "ObjectMapper\|JsonInclude\|default-property-inclusion" src/main` and `application.yml` both come back empty. The body this handler returns, captured from a real run:

```json
{"status":400,"message":"Malformed request body","path":"/api/upgrades","timestamp":[...],"fieldErrors":null}
```

Every 400, 404, 409, 422 and 500 the API returns carries `"fieldErrors":null`. Either add `@JsonInclude(JsonInclude.Include.NON_NULL)` to `ErrorResponse` — which is what the comment claims is happening — or correct the comment.

backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
NaimElijah, line 182: **The one log line that matters for diagnosis cannot be correlated with the client's report.**

`~/.claude/CLAUDE.md`: *"Propagate a correlation or trace id across every boundary and include it in every log line."* `grep -rn "MDC\|correlationId\|traceId\|X-Request-Id" src` returns nothing — the repo has no correlation infrastructure at all.

This is the line where its absence costs most. A user reports "I got a 500", and the only join key between their `ErrorResponse` and this log entry is `path` plus a `LocalDateTime.now()` with no zone, on a server whose clock the reporter cannot see. Under any concurrency that is not a join key.

Since the PR's own argument is that a withheld cause is only defensible if the server keeps it, the same argument applies to being able to *find* what the server kept: put a request id in the MDC (a one-filter change) and return it in `ErrorResponse`. If that is out of scope here, it belongs in Open questions rather than being silently absent.

backend/src/test/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandlerTest.java
```diff
@@ -90,6 +118,39 @@ void accessDenied_isForbidden() {
         assertThat(response.getStatusCode()).isEqualTo(HttpStatus.FORBIDDEN);
     }
 
+    @Test
+    void unknownEnumValueInBody_isBadRequest() throws Exception {
```
NaimElijah: **Test names here break the naming rule in `~/.claude/CLAUDE.md`.**

*"Tests are the executable specification. Name them like so: `Given<input/scenario>_When<function/process>_Then<result>`."*

The four new tests are `unknownEnumValueInBody_isBadRequest`, `malformedJsonBody_isBadRequest` (l.132), `unbindableBody_saysNothingAboutTheParser` (l.140) and `unexpectedFailure_isLoggedWithItsStackTrace` (l.167) — two-part names with no `When`. They match the existing style in this file, so this is consistency-with-neighbours versus the stated global rule; flagging it so the choice is deliberate rather than accidental. `GivenUnknownEnumConstantInBody_WhenPostingAnUpgrade_ThenBadRequest` is the shape the rule asks for.

backend/src/test/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandlerTest.java
```diff
@@ -1,37 +1,65 @@
 package com.healthupgrades.common.adapter.in.web;
 
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
```
NaimElijah, line 6: **`ch.qos.logback` is imported directly but declared nowhere.**

`ListAppender`, `Level`, `Logger` and `ILoggingEvent` reach the test classpath only transitively: `spring-boot-starter-web` → `spring-boot-starter` → `spring-boot-starter-logging` → `logback-classic`. `backend/pom.xml` names no logback dependency.

It compiles today. It stops compiling the day someone swaps the logging starter (log4j2 is the usual reason), and the failure will point at a test that has nothing to do with the change. `~/.claude/CLAUDE.md` asks for dependencies to be pinned and justified; a direct `import` of a library the POM does not name is the case that rule exists for. Add an explicit `ch.qos.logback:logback-classic` with `<scope>test</scope>`, version-managed by the Boot parent.

backend/CLAUDE.md
```diff
@@ -53,7 +53,13 @@ side. Both are deliberate, not drift.
 - **Exception → HTTP status is centralized** in the global exception handler. Throw the
   right domain exception rather than building `ResponseEntity` status by hand:
   `ResourceNotFoundException` → 404, `BusinessRuleException` → 422, `DuplicateProgressException` /
-  optimistic-lock → 409, bean-validation → 400.
+  optimistic-lock → 409, bean-validation and an unbindable body (`HttpMessageNotReadableException`)
+  → 400.
+
+  The advice declares a catch-all `@ExceptionHandler(Exception.class)`, and an advice is consulted
+  before Spring's own `DefaultHandlerExceptionResolver` — so any framework exception without an
```
NaimElijah: **This paragraph documents the trap accurately and then prescribes the wrong remedy.**

*"Add the handler when you meet one; that is what issue #22 was."*

"Meet one" means: a user hits it in production, gets a 500, and someone traces it back. The diagnosis above it is right — a catch-all in an advice pre-empts `DefaultHandlerExceptionResolver` — but the conclusion should be to remove the pre-emption, not to enumerate its victims one incident at a time. Four are already live and unhandled on `UpgradeController` alone (bad UUID, bad enum query param, wrong method, wrong media type; all verified returning 500).

Once `handleGeneral` honours a framework exception's own status, this guidance becomes shorter and more useful: *"the catch-all defers to Spring's mapping; add a handler only when you want a status different from the framework's."*

NaimElijah:
## Review response — 37d18b6

The central finding was right and I verified it independently before acting on it. Against the first
commit, on `UpgradeController` alone:

```
GET   /api/upgrades/not-a-uuid     -> 500   (should be 400)
GET   /api/upgrades?status=BOGUS   -> 500   (should be 400)
PATCH /api/upgrades/{id}           -> 500   (should be 405)
POST  /api/upgrades  text/plain    -> 500   (should be 415)
```

The first commit fixed one tile of a whack-a-mole and its own documentation told the next person to
keep playing it. `GlobalExceptionHandler` now extends `ResponseEntityExceptionHandler`, so Spring's
mapping applies to every framework exception and `handleExceptionInternal` re-clothes the result in
this API's `ErrorResponse` body. Reasoning, the rejected alternatives and the revisit triggers are in
[ADR-006](docs/ADRs/ADR-006-framework-exceptions-through-responseentityexceptionhandler.md).

### Fixed

| # | Finding | What changed |
|---|---|---|
| 1 | Catch-all still steals every other framework exception | Extends `ResponseEntityExceptionHandler`; four verified 500s now 400/400/405/415, and an unmatched route is 404 instead of 500 |
| 2 | ERROR + stack trace for client-caused failures | 4xx logs at DEBUG, 5xx at ERROR. Pinned by `GivenAClientMistake_WhenItIsRejected_ThenNothingIsLoggedAtError` |
| 4 | `handleNotReadable` logged nothing | Every framework rejection now logs with its exception through one exit; an I/O failure reading a body leaves a trace |
| 5 | NFR-7 asserted a contract the code did not honour, and cited a test that pinned nothing | NFR-7 rewritten to what is now true; bean-validation is covered by a test for the first time |
| 6 | `saysNothingAboutTheParser` passed against the unfixed code | Now asserts the status, the message, the path and the absence of `fieldErrors` |
| 8 | `"fieldErrors":null` shipped despite the javadoc | `@JsonInclude(NON_NULL)` on `ErrorResponse` — the javadoc was the intent, the annotation was missing |
| 10 | Test names | The tests this PR adds use `Given…_When…_Then…`. The seven pre-existing ones are left alone: renaming them is a separate change from a bug fix |
| 12 | The guidance institutionalised whack-a-mole | Replaced. Override the `handleXxx` hook; a second `@ExceptionHandler` for a type the parent maps is an ambiguous mapping that fails at boot — which `GivenTheAdviceAsSpringSeesIt_WhenItsHandlerMappingsAreBuilt_ThenNoTypeIsMappedTwice` now catches in `mvn test` |

### Partly accepted

**#3 — logging the exception verbatim vs NFR-6.** The tension is real, the proposed direction is not:
a 500 whose cause is withheld from the client and not kept by the server is undiagnosable, which is
what #22 asked to fix. Two things changed instead. The stack trace now fires only for a genuine 5xx,
not for the client mistakes the catch-all used to claim. And NFR-6 no longer claims something false —
it states the exception explicitly rather than being quietly contradicted by the code under it.

**#7 — the MockMvc rig.** The rig moved into a `@Nested` class, so the eight direct-call tests no
longer build a dispatcher and two mocks they never touch. The `ObjectMapper` difference is real and
now documented on the class: the rig pins statuses and the error body's own fields, not how a domain
type serialises — `UpgradeDtoSerializationTest` owns that. I kept the real controller deliberately: a
fixture controller would test that the advice works in front of a controller I wrote for the test,
which is the part that was never in doubt.

### Not accepted

**#11 — logback imported but not declared in `pom.xml`.** Declaring it would pin a transitive of
`spring-boot-starter-logging`, which is the starter's own default and not an incidental dependency;
the failure mode described (swapping to log4j2) would break far more than this test and is diagnosable
in one line. The alternative — asserting on captured console text — tests the log pattern instead of
the log event. `ListAppender` asserts the level and the attached throwable, which is what the
requirement is about.

### Split out

**#9 — no correlation id.** Confirmed: `grep -rn 'MDC|correlationId|traceId|X-Request-Id' backend/src`
returns nothing, so a reported 500 cannot be joined to its log line. That is a cross-cutting concern
with its own design (filter, MDC lifecycle, response header, and a way in for WebSocket and the
schedulers), not part of a status mapping. Filed as #43.

### Verification

`mvn test` — 155 tests, 0 failures (was 148; the exception contract went from 8 assertions to 18).
Every status quoted above is asserted, not described.

PR #44, opened 2026-08-23
feat(obs): give every log line, response and job a trace id

### LINKED ISSUE #43: obs: no correlation id, so a reported 500 cannot be joined to its log line
Split out of the review on #42, which found it while fixing the exception → status contract. Not
fixed there: it is a cross-cutting concern with its own design, not part of a status mapping.

## Verified

`grep -rn 'MDC|correlationId|traceId|X-Request-Id' backend/src` returns nothing. There is no
correlation id anywhere in the backend: not in the log pattern, not on the response, not propagated
across a boundary.

## Impact

`GlobalExceptionHandler` logs an unexpected 5xx at ERROR with its stack trace, and returns a body
carrying only `status`, `message`, `path` and a zone-less `LocalDateTime.now()`. A user reporting
"I got a 500 on /api/upgrades" gives no key that identifies their log line: under concurrent traffic
on the same endpoint, several entries match equally well. The trace exists and cannot be found.

The same gap applies to every log line, not only errors — the global convention is that a correlation
id is propagated across every boundary and included in every line, and today none is.

## Sketch

- A servlet filter that reads an inbound `X-Request-Id` (or mints one), puts it in the SLF4J `MDC`,
  clears it in a `finally`, and echoes it on the response.
- The id included in the logging pattern, so every line carries it without a call site knowing.
- `ErrorResponse` gains the id, so a screenshot of a failure is enough to find the log entry.
- WebSocket and the scheduled jobs need their own way in — neither has a request to read a header
  from, and both log today.

## Acceptance criteria

- [ ] Every log line emitted while serving a request carries the same id
- [ ] The id is returned to the client on an error response and echoed as a response header
- [ ] An inbound id is honoured rather than replaced, so a caller can correlate across services
- [ ] The MDC is cleared at the end of the request, verified by a test that a second request on the
      same thread does not inherit the first one's id

NaimElijah:
Closes #43.

## Problem

[#42](https://github.com/NaimElijah/UpHealther/pull/42) made the advice log an unexpected 5xx at ERROR
with its stack trace. That improved nothing a user could act on: the error body carries `status`,
`message`, `path` and a zone-less timestamp, so "I got a 500 on `/api/upgrades`" matches every 500 that
endpoint ever produced. The trace existed and could not be found.

Nothing in the backend emitted a correlation id at all — no MDC key, no response header, no field on any
response. That included the DEBUG line `SpringDomainEventPublisher` writes for every domain event, which
is the only thing the scheduled jobs log and which fires every minute from `dispatchReminders`.

## Approach

Micrometer Tracing with the OpenTelemetry bridge, per the "instrument vendor-neutrally" standard, rather
than the hand-rolled `X-Request-Id` filter the issue sketched. Recorded in
[ADR-007](docs/ADRs/ADR-007-request-correlation-through-micrometer-tracing.md) with the alternatives and
what would make each worth revisiting.

Most of the behaviour falls out of the dependency, which is why the diff is smaller than the scope
suggests — all three verified against the jars rather than assumed:

- `LogCorrelationEnvironmentPostProcessor` plus Boot's `logback/defaults.xml`, which already interpolates
  `${LOG_CORRELATION_PATTERN}`, put the id on every log line with **no `logback-spring.xml` and no
  `logging.pattern.*`**.
- Inbound W3C `traceparent` is consumed by default, so a caller's trace is continued.
- `ServerHttpObservationFilter` opens the scope in a try-with-resources, so the MDC clears when the
  request ends whether or not it failed.

Four things had to be written:

| What | Why it wasn't free |
|---|---|
| `TraceIdResponseHeaderFilter` → `X-Trace-Id` | Set **before** the chain, so it survives `sendError` and reaches the rejection Spring Security produces without ever entering `GlobalExceptionHandler`. Registered explicitly at `HIGHEST_PRECEDENCE + 2`: after the observation filter that opens the scope, before Spring Security at `-100`. A `@Component` filter defaults to `LOWEST_PRECEDENCE`, on the wrong side of both. |
| `traceId` on `ErrorResponse` | The body is what a user screenshots. Routed through one private `body(...)` factory instead of widening the 3-arg constructor across 8 call sites, so a future handler cannot forget it. |
| `ObservabilityConfig implements SchedulingConfigurer` | Spring 6.1 already wraps every `@Scheduled` run in an observation, but Boot 3.2.5 never supplies the registry. One line covers all three jobs and any added later — no scheduler class is touched. |
| `StompTracingChannelInterceptor` | `ExecutorSubscribableChannel.sendInternal` only *dispatches*, so `preSend`/`afterSendCompletion` bracket the dispatch, not the handling. Both threads need a scope, joined by the sending span's `TraceContext` on a non-native header. Registered **before** `JwtChannelInterceptor`, because that one rejects a bad CONNECT by throwing from `preSend` and the chain unwinds in reverse. |

## Trade-off accepted

Correlation is now a property of the runtime rather than of each call site, so part of the behaviour
lives in auto-configuration whose defaults can change with a Boot upgrade. `RequestCorrelationTest`
boots the real auto-configuration so that such a change fails a test rather than quietly
un-correlating the logs.

Sampling is `1.0` with no exporter. The id reaches the MDC either way — Micrometer writes it without
consulting the sampler — but an unsampled span silently drops every tag. ADR-007 records that this is
the volume that would start leaving the process if a collector is ever added.

## Verification

`mvn verify` against a throwaway Postgres — **172 unit tests, 4 integration tests, BUILD SUCCESS**.
`mvn test` remains database-free, as `backend/CLAUDE.md` requires: the MDC-lifecycle test boots Boot's
tracing auto-configuration with no web server or datasource.

Two tests were wrong in ways only running them exposed, both now fixed and both worth knowing about:

- **`CorrelationIT` was measuring a context production never runs.** `@SpringBootTest` sets
  `management.tracing.enabled=false` unless `@AutoConfigureObservability` is present. The tracer still
  mints ids, so `X-Trace-Id` looked healthy, but the propagator resolves to `NoopTextMapPropagator` with
  no fields — nothing is ever extracted. Confirmed from the condition-evaluation report, not inferred.
  Separately, `TestRestTemplate` is instrumented and injects its own `traceparent` over the test's, so
  the IT now uses the JDK's uninstrumented `HttpClient`.
- **The STOMP continuation test passed with the parent link deleted.** Handling on the sending thread
  inherits the ambient span, so the assertion held however the parent was wired. It now runs the handler
  half on its own thread, as production does; deleting `setParent` fails it.

## Limits, stated rather than discovered later

1. **`X-Request-Id` is not honoured.** An OTel trace id is 32 lowercase hex characters; a dashed UUID or
   an nginx `$request_id` cannot become one. AC 3 is met in substance via `traceparent`, not in the
   header the issue named. Worth confirming no proxy in front of this service sets one.
2. **An anonymous request carries the header but not the body field** — it is rejected inside the
   security chain and returns Boot's default error body. It is also a **403, not a 401**: with no
   `AuthenticationEntryPoint` configured, `Http403ForbiddenEntryPoint` answers. Worth knowing before
   anyone writes a client that branches on 401. Its own issue.
3. **`ServerHttpObservationFilter` covers `REQUEST` and `ASYNC` dispatches, not `ERROR`**, so a container
   error dispatch to `/error` has no id. Nothing logs there today.
4. **`spring.application.name` now prefixes every log line** (a separate Boot converter from the
   correlation field). It is set because it also becomes OTel's `service.name`; set
   `logging.include-application-name: false` if the column is unwanted.
5. **A `BatchSpanProcessor` daemon thread runs with no exporter.** Harmless, and it must not be
   suppressed by overriding the `SpanProcessors` bean — Boot collects it, so an empty override leaves the
   thread running *and* unwires it.
6. **The frontend surfaces none of this.** `api/client.ts` branches only on 401 and reads no error body
   or header. Showing the id to a user is a separate change.

---

**Post-review update (`59be5ff`).** Fourteen findings: twelve fixed, one documented (an inbound
`traceparent` is caller-controlled — inherent to honouring it, now in ADR-007 with a revisit trigger),
one declined with reasons. The one that mattered was a scope leak: `preSend` opened the tracing scope
before a mutation that can throw, and Spring skips `afterSendCompletion` for an interceptor whose own
`preSend` threw — leaving an attached scope on a pooled thread that would stamp a dead frame's trace id
on every later frame. Full response in the [review reply](https://github.com/NaimElijah/UpHealther/pull/44#issuecomment-5388959136).

Verified after the fixes: **175 unit tests, 5 integration tests, BUILD SUCCESS**, and CI green.


backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+    private final Propagator propagator;
+
+    @Override
+    public Message<?> preSend(Message<?> message, MessageChannel channel) {
+        StompHeaderAccessor accessor = StompHeaderAccessor.wrap(message);
+        Span span = propagator
+                .extract(accessor, (carrier, key) -> carrier.getFirstNativeHeader(key))
+                .name(spanName(accessor))
+                .start();
+        open(span);
```
NaimElijah: The span is opened before the two statements that can still throw (`getMutableAccessor` / `setHeader` on an accessor another interceptor already marked immutable, `MessageBuilder.createMessage` on a null payload). If `preSend` throws after this line, Spring never calls this interceptor's `afterSendCompletion`: `ChannelInterceptorChain.applyPreSend` only advances `sendInterceptorIndex` **after** a successful `preSend`, and `triggerAfterSendCompletion` iterates from that index down â€” so the index is still `-1` and this frame is skipped.

The consequence is not a lost span, it is a leaked one: the `Tracer.SpanInScope` is never closed, so the OTel context stays attached to a **pooled** channel-executor thread and every later frame handled on that thread reports the previous frame's trace id.

Build the message first and call `open(span)` last, or wrap the body in `try { ... } catch (RuntimeException e) { close(e); throw e; }`.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+public class StompTracingChannelInterceptor implements ExecutorChannelInterceptor {
+
+    /** Internal header carrying the sending span's context to the handling thread. Never sent to a client. */
+    private static final String TRACE_CONTEXT_HEADER = "healthupgrades-trace-context";
+
+    /**
+     * Spans opened on this thread and not yet closed. A deque rather than a single slot because a
+     * handler may itself send on another channel, nesting a second span inside the first.
+     */
+    private static final ThreadLocal<Deque<Scope>> OPEN_SCOPES = ThreadLocal.withInitial(ArrayDeque::new);
```
NaimElijah: `OPEN_SCOPES` is `static`, so every instance of this interceptor â€” and every test that constructs one â€” shares one deque per thread. The class is a singleton `@Component`, so `static` buys nothing, and it actively costs correctness in the tests: `StompTracingChannelInterceptorTest.GivenTheSpanContextTravels_WhenTheFrameIsEncoded_ThenItIsNotANativeHeader` calls `preSend` with no matching `afterSendCompletion`, leaving a `Scope` on the JUnit thread's deque that a *different* test's interceptor instance then pops.

Making it an instance field scopes the deque to the interceptor that opened the spans and removes the cross-test coupling.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+     * one's {@code preSend} ran, and Spring still calls the completion hook.
+     */
+    private void close(Exception ex) {
+        Deque<Scope> scopes = OPEN_SCOPES.get();
+        Scope scope = scopes.poll();
+        if (scope == null) {
+            return;
+        }
+        if (scopes.isEmpty()) {
+            OPEN_SCOPES.remove();
```
NaimElijah: The early return above (`scope == null`) skips `OPEN_SCOPES.remove()`, so the `ThreadLocal.withInitial` deque created by `OPEN_SCOPES.get()` on line 104 stays attached to that thread for good. That is exactly the path the javadoc documents as normal ("an interceptor earlier in the chain can reject a message before this one's `preSend` ran"), and it happens on pooled container/executor threads.

Removing the empty deque in the `scope == null` branch too keeps the thread-local map clean.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public void afterSendCompletion(Message<?> message, MessageChannel channel, boolean sent, Exception ex) {
+        close(ex);
+    }
+
+    @Override
+    public Message<?> beforeHandle(Message<?> message, MessageChannel channel, MessageHandler handler) {
+        Span.Builder builder = tracer.spanBuilder().name(spanName(StompHeaderAccessor.wrap(message)) + " handle");
+        if (message.getHeaders().get(TRACE_CONTEXT_HEADER) instanceof TraceContext parent) {
+            builder.setParent(parent);
```
NaimElijah: The return value of `setParent` is discarded. `Span.Builder` is a fluent interface: `OtelSpanBuilder.setParent` happens to mutate and return `this`, so this works with the bridge wired today, but the contract does not require it â€” a builder that returns a new instance would silently drop the parent and start an unrelated root span on the handling thread. That is precisely the failure `GivenTheHandlerRunsOnAnotherThread_...` was written to catch, and it would not catch it, because the test asserts through the same OTel bridge that makes the shortcut work.

`builder = builder.setParent(parent);` costs nothing and removes the assumption.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+            scope.span.error(ex);
+        }
+        scope.inScope.close();
+        scope.span.end();
+    }
+
+    /** {@code SEND /app/x}, or just the command when the frame has no destination. */
+    private static String spanName(StompHeaderAccessor accessor) {
+        String command = accessor.getCommand() == null ? "MESSAGE" : accessor.getCommand().name();
+        return accessor.getDestination() == null ? command : command + " " + accessor.getDestination();
```
NaimElijah: The span name is `COMMAND + " " + destination`, destination verbatim. On the broker channel the first hop of `convertAndSendToUser` carries `/user/<userId>/queue/notifications`, so **a user id ends up in the span name**; the second hop carries `/queue/notifications-user<sessionId>`. Unbounded cardinality either way.

This PR also amends NFR-6 to say "a trace id identifies a request rather than a person"; a span name that embeds the user id is the counter-example. Nothing exports today, but ADR-007 records adding a collector as the expected next step, at which point this leaves the process.

Either template the user segment out of the name (`SEND /user/{userId}/queue/notifications`) or name the span after the command alone and keep the destination as a tag.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    private static final ThreadLocal<Deque<Scope>> OPEN_SCOPES = ThreadLocal.withInitial(ArrayDeque::new);
+
+    private final Tracer tracer;
+    private final Propagator propagator;
+
+    @Override
+    public Message<?> preSend(Message<?> message, MessageChannel channel) {
+        StompHeaderAccessor accessor = StompHeaderAccessor.wrap(message);
+        Span span = propagator
```
NaimElijah, line 68: The `traceparent` is taken from client-supplied STOMP native headers, and the HTTP side does the same with the `traceparent` request header â€” on an API whose `/api/auth/**` and `/ws/**` endpoints are `permitAll`. There is no trusted-proxy boundary, so any unauthenticated caller can pin every one of their requests to a single trace id, or reuse an id they read out of someone else's `X-Trace-Id`. The log search this PR exists to enable then returns two users' lines under one id.

ADR-007 presents "inbound `traceparent` is consumed by default" as a pure win. Worth recording the trust assumption next to it, and deciding whether inbound trace context should be accepted only from a known proxy.

backend/src/test/java/com/healthupgrades/common/observability/StompTracingChannelInterceptorTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+            assertThat(sendingTraceId).isNotBlank();
+            assertThat(handlingTraceId).isEqualTo(sendingTraceId);
+        });
+    }
+
+    @Test
+    void GivenTheSpanContextTravels_WhenTheFrameIsEncoded_ThenItIsNotANativeHeader() {
+        // A native header would be written to the wire by StompEncoder and shipped to the browser.
+        Message<?> forwarded = interceptor.preSend(frame(StompCommand.SEND, "/app/ping"), channel);
```
NaimElijah, line 99: This test opens a span and never closes it â€” no `afterSendCompletion`. Because `OPEN_SCOPES` is `static`, the `Scope` stays on the JUnit thread's deque for the rest of the fork.

That makes `GivenAnInterceptorAheadOfThisOneRejectedTheFrame_WhenCompletionRuns_ThenNothingBlowsUp` (line 120) order-dependent and, in the common ordering, **vacuous**: instead of reaching the `scope == null` branch it exists to cover, `close()` pops this leaked scope. Its assertion still passes only because `tracer` is a fresh `SimpleTracer` per test method, so the popped span belongs to a different tracer and never appears in `getSpans()`.

Close the span here (`interceptor.afterSendCompletion(forwarded, channel, true, null)`), or make the deque an instance field â€” which fixes this and the thread-local leak together.

backend/src/test/java/com/healthupgrades/common/observability/RequestCorrelationTest.java
```diff
@@ -0,0 +1,132 @@
...
+ * one's id" a deterministic assertion rather than a race. Against a real Tomcat the thread cannot be
+ * pinned, so {@code CorrelationIT} checks the wire contract and this checks the MDC lifecycle.
+ *
+ * <p>The guarantee itself belongs to the framework: {@code ServerHttpObservationFilter.doFilterInternal}
+ * opens the observation scope in a try-with-resources, so the scope — and with it the MDC — closes even
+ * when the chain throws.
+ *
+ * <p>One caveat worth knowing if this ever fails oddly: Boot installs the MDC bridge by calling
+ * {@code ContextStorage.addWrapper(...)}, a JVM-global side effect that OpenTelemetry ignores once a
+ * {@code Context} has been used. That holds while this is the only test touching OTel context in the
```
NaimElijah: This caveat stopped being true in this PR. `StompTracingChannelInterceptorTest` (line 43) boots the same `OpenTelemetryAutoConfiguration` in the same Surefire fork â€” the build sets no `forkCount`/`reuseForks`, so it is one JVM for every unit test.

Boot registers the MDC bridge in `otelCurrentTraceContext(...)` via `ContextStorage.addWrapper(new EventPublishingContextWrapper(publisher))`, and OpenTelemetry ignores every `addWrapper` after the storage has been initialised. Whichever of these two classes touches an OTel `Context` first therefore wins, and this class's MDC assertions end up riding a wrapper bound to the other test's (already closed) application context.

Either update the comment to name the second class and say why it is still safe, or make the isolation explicit rather than relying on being alone in the fork.

backend/src/main/java/com/healthupgrades/common/observability/ObservabilityConfig.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * <p>Spring already does the work — {@code ScheduledMethodRunnable} wraps each invocation in an
+     * observation and closes the scope in a {@code finally} — but it needs a registry, and Boot 3.2.5
+     * never supplies one. This one line covers all three jobs and any job added later, which is why the
+     * scheduler classes themselves are untouched: nothing for a new job to remember, and no
+     * {@code try/finally} for anyone to get wrong.
+     */
+    @Override
+    public void configureTasks(ScheduledTaskRegistrar registrar) {
+        registrar.setObservationRegistry(observationRegistry);
```
NaimElijah, line 59: Nothing tests this line, and it is the entire scheduled-job half of NFR-21 ("every log line written while ... running a scheduled job ... carries the same trace id"). Delete it and the build stays green â€” `mvn test` and `mvn verify` both pass â€” while all three jobs silently stop being correlated, including `dispatchReminders`, which ADR-007 names as the once-a-minute case that motivated this.

The same gap covers the ordering the other bean in this class is built around: the 401-from-the-security-chain case that justifies `HIGHEST_PRECEDENCE + 2` is asserted only against a `MockFilterChain` calling `sendError`, never against the real chain â€” `CorrelationIT` exercises 200 and 400 only.

A test that drives one `@Scheduled` method through a registrar configured by this class, plus a `CorrelationIT` case hitting an authenticated endpoint with no token, would close both.

backend/src/main/java/com/healthupgrades/common/adapter/in/web/GlobalExceptionHandler.java
```diff
@@ -222,6 +244,7 @@ public static class ErrorResponse {
         private String path;
         private LocalDateTime timestamp;
         private Map<String, String> fieldErrors;
+        private String traceId;
 
         /**
```
NaimElijah, line 249: The PR describes routing every body through the private `body(...)` factory "so a future handler cannot forget it", but `ErrorResponse` is a `public static` class with a `public` 3-arg constructor. `new ErrorResponse(status, message, path)` still compiles anywhere and yields a body with no `traceId` â€” the guarantee is a convention, not a constraint.

Making the constructor `private` (its only caller is `body(...)`, line 90) turns the intent into something the compiler enforces.

backend/src/test/java/com/healthupgrades/CorrelationIT.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+@AutoConfigureObservability
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class CorrelationIT {
+
+    private static final String TRACE_ID_HEADER = "X-Trace-Id";
+    private static final String INBOUND_TRACE_ID = "4bf92f3577b34da6a3ce929d0e0e4736";
+    private static final String TRACE_ID_PATTERN = "[0-9a-f]{32}";
+
+    private final HttpClient client = HttpClient.newHttpClient();
```
NaimElijah: `HttpClient` implements `AutoCloseable` as of Java 21 (this module targets 21), and this one is never closed. JUnit's default per-method lifecycle means one client per test â€” three here â€” each holding a selector thread and an executor for the rest of the fork.

Either make it `static final` so there is one, or close it in an `@AfterEach`.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+                .start();
+        open(span);
+
+        MessageHeaderAccessor mutable = MessageHeaderAccessor.getMutableAccessor(message);
+        mutable.setHeader(TRACE_CONTEXT_HEADER, span.context());
+        return MessageBuilder.createMessage(message.getPayload(), mutable.getMessageHeaders());
+    }
+
+    @Override
+    public void afterSendCompletion(Message<?> message, MessageChannel channel, boolean sent, Exception ex) {
```
NaimElijah, line 90: `sent` is ignored, so a send the channel refused â€” `ExecutorSubscribableChannel.send` returning `false` on a timeout, or a broker channel with no subscriber â€” ends the span as an unremarkable success with `ex == null`. That is the case where a push silently did not happen, which is the one worth finding in a trace afterwards.

Recording it (`if (!sent) scope.span.tag(...)`, or an event) costs one line and makes the span say what actually happened.

backend/src/main/java/com/healthupgrades/common/observability/StompTracingChannelInterceptor.java
```diff
@@ -0,0 +1,127 @@
...
+        StompHeaderAccessor accessor = StompHeaderAccessor.wrap(message);
+        Span span = propagator
+                .extract(accessor, (carrier, key) -> carrier.getFirstNativeHeader(key))
+                .name(spanName(accessor))
+                .start();
+        open(span);
+
+        MessageHeaderAccessor mutable = MessageHeaderAccessor.getMutableAccessor(message);
+        mutable.setHeader(TRACE_CONTEXT_HEADER, span.context());
+        return MessageBuilder.createMessage(message.getPayload(), mutable.getMessageHeaders());
```
NaimElijah: Every frame on all three channels gets a fresh `Message` allocation plus a header write, including every outbound push to every subscriber. The header only earns its keep where the handling thread differs from the sending thread. On the broker channel it does not: `AbstractMessageBrokerConfiguration` gives that channel no executor unless one is registered, so dispatch is synchronous and `beforeHandle` runs on the thread that already has the span in scope â€” where `OtelPropagator.extract` would find the parent through `Context.current()` anyway.

Attaching the context header only when the channel dispatches asynchronously would keep the hot push path allocation-free.

backend/src/main/java/com/healthupgrades/common/observability/ObservabilityConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.Ordered;
+import org.springframework.scheduling.annotation.SchedulingConfigurer;
+import org.springframework.scheduling.config.ScheduledTaskRegistrar;
+
+/**
+ * Wires the two pieces of correlation that Spring Boot does not wire itself.
```
NaimElijah: "the two pieces" undercounts: this PR wires three things Boot does not â€” the response-header filter, the scheduled-task registry, and `StompTracingChannelInterceptor`, which is a `@Component` registered from `WebSocketConfig` rather than from here. A reader who trusts this sentence will not go looking for the STOMP half.

Either say "the pieces ... that live here" and point at `WebSocketConfig`, or move the STOMP registration into this class so the count is right.

NaimElijah:
## Review response — 59be5ff

Twelve of fourteen findings fixed, one documented, one declined. The most valuable one was the scope
leak; the review also contains one claim that is demonstrably wrong, noted at the end.

### Fixed

| # | Finding | What changed |
|---|---|---|
| 1 | Scope opened before a mutation that can throw | The real bug. `ChannelInterceptorChain` advances its index only after `preSend` returns, so an interceptor that throws is skipped by `triggerAfterSendCompletion` — leaving an attached scope on a pooled thread that stamps the dead frame's id on every later frame. The scope now opens once nothing above it can throw, and a failure ends the span on the way out. |
| 2 | `static` ThreadLocal shared across instances | Instance field. It is a singleton, so `static` bought nothing and braided every test-constructed instance into the same deque. |
| 3 | A test left a scope open, making the next one vacuous | That test now completes its send. With #2 fixed the two are independent regardless. |
| 4 | `RequestCorrelationTest` javadoc claimed to be the only test touching OTel context | It was true when written and stopped being true inside the same PR. Reworded to say what actually holds and why it is safe. |
| 5 | No coverage of the scheduled-job wiring or the security-rejection path | `ObservabilityConfigTest` pins both one-liners whose absence is silent, and `CorrelationIT` now covers the rejection path. |
| 6 | Span names carried the destination, so a user id | `convertAndSendToUser` produces `/user/{userId}/queue/notifications`. Names are the command only — it contradicted this PR's own NFR-6 amendment and gave unbounded cardinality. |
| 8 | Empty deque retained on the thread | `remove()` now runs on that path too. |
| 9 | `setParent`'s return value discarded | Used rather than assumed to be `this`. |
| 10 | `ErrorResponse` constructor still public | Private. "A future handler cannot forget the trace id" was a convention; `body(...)` being the only way to make one is what makes it a constraint. |
| 11 | `HttpClient` never closed | One per class, closed in `@AfterAll`. |
| 12 | A refused send ended as a success | Tagged `stomp.send.refused`, with a test. |
| 14 | Javadoc said "two pieces", the PR wires three | Corrected, and it now points at `WebSocketConfig` for the third. |

### Documented, not changed

**#7 — inbound `traceparent` is caller-controlled.** Correct, and inherent to honouring an inbound id at
all, which is the issue's third acceptance criterion. Anyone can pin requests to an id of their choosing.
The mitigation is a trusted-proxy boundary, which does not exist to configure against yet. Recorded in
ADR-007 with the revisit trigger, plus the line that matters: a trace id is a correlation hint, never
evidence, and is not an authorization input anywhere.

### Declined

**#13 — per-frame allocation on the synchronous broker channel.** Real but not worth the conditional. The
saving is one header write and one message rebuild on a channel that is not hot, and the branch would
have to encode which channels have executors — a fact that changes if anyone registers one, silently
breaking correlation rather than performance. Revisit if the broker channel ever shows up in a profile.

### One finding in the review is wrong

The verification note states *"`SimpleTracer.nextSpan()` does generate a trace id, so the new
`GlobalExceptionHandlerTest`/`CorrelationIdTest` assertions hold."* It does not. CI failed on exactly
that — `expected: "" but was: null` — because the trace id is blank and `CorrelationId` correctly rejects
a blank id. Fixed in `a4c9d2e`, before this review landed: both tests now set an explicit id and assert
against that constant, so they cannot pass by comparing two blank strings.

### One thing the fixes turned up

Adding the security-rejection test showed an anonymous request returns **403, not 401**. With no
`AuthenticationEntryPoint` configured, Spring Security leaves `Http403ForbiddenEntryPoint` in place. My
PR description, ADR-007, `architecture.md` and two javadocs all said 401 — corrected everywhere, since
anyone writing a client that branches on 401 would have been misled by my own documentation.

### Verification

`mvn verify` against a throwaway Postgres: **175 unit tests, 5 integration tests, BUILD SUCCESS.**

PR #47, opened 2026-08-24
test: make the suite cover what requirements.md claims it enforces

### LINKED ISSUE #46: Test: Enhance/Expand whole test suite
`docs/requirements/requirements.md` opens with a promise:

> where it is enforced in code the enforcing class or test is named, so a claim here can be checked rather than trusted.

Today most entries name a **class**, not a test. So most of the document is trusted, not checked.

## The audit

| Area | State |
|---|---|
| `auth` + `common/security` | **No tests at all.** FR-1…FR-5, NFR-1…NFR-5 rest on inspection. |
| `healtharea`, `reflection`, `reminder`, `user` services | **No tests at all.** FR-6…FR-8, FR-23…FR-26, BR-13. |
| All 9 controllers | **No web-layer test exists.** FR-5 (auth required) and BR-15 (foreign row → 404) are asserted nowhere. |
| All 8 persistence adapters | **No test exists.** BR-6 (unique constraint) and BR-14 (`@Version` → 409) are database behaviour no unit test can see. |
| `TrackingService` (241 lines) | 2 tests. FR-20, FR-21 and BR-10 uncovered. |
| `DashboardAggregationService` | 1 test for the whole of FR-27 — the context that already shipped #20. |
| Frontend `client.ts`, `AuthContext`, `ProtectedRoute`, `NotificationProvider` | **No tests.** The 401→logout path and session restore (FR-3) are unasserted. |
| Naming | Backend uses `method_scenario_shouldResult`; the project standard is `Given…_When…_Then…`. |
| `mvn verify` | Needs a Postgres the developer wires up by hand, so it can silently run against the wrong database. |
| Coverage | JaCoCo runs and the report is surfaced nowhere. The frontend has no coverage provider at all. |

## What this closes

Two things: the tests that already exist are brought onto one convention and one set of levels, and the
gaps above are closed so `requirements.md` can name a test for every row that can name one.

## Decisions taken

- **Scope** — backend and frontend, full sweep.
- **Integration-test database** — Testcontainers, so `mvn verify` starts its own throwaway Postgres and
  cannot hit the wrong one. Requires Docker. → ADR-008.
- **Real-browser tests** — deferred again. FR-39, NFR-19 and NFR-20 stay hand-checked and stay in §6 of
  the requirements. Playwright drags the backend, a database and a browser binary into CI; that is its
  own decision, not a rider on this one.
- **Naming** — one mechanical rename of the existing suite to `Given…_When…_Then…`, then every new test
  follows it. → ADR-009.

## Definition of done

Every FR / BR / NFR either names a test that enforces it, or is listed in §6 as a deliberate, explained
gap.

Explicitly **not** a coverage percentage: per the project standard, coverage is reported and never gated.

## Out of scope

- Playwright, axe, or any real-browser gate.
- Mutation testing (PIT).
- The backend NVD dependency audit — ADR-002's open item, needs a repository secret.
- Production-code changes, unless a test proves a genuine bug. Such a fix gets its own `fix:` commit with
  the failing test first, and is called out rather than folded in quietly.

NaimElijah:
Closes #46.

## The problem

`docs/requirements/requirements.md` opens by promising that "the enforcing class or test is named, so a
claim here can be checked rather than trusted". Most rows named a **class**, which is not checkable — a
class exists whether or not anything asserts what the row claims about it.

Behind that, the audit found whole areas with no tests at all: the `auth` context and `common/security`,
the `healtharea`, `reflection`, `reminder` and `user` services, all nine controllers, and all eight
persistence adapters. On the frontend, the axios client, the auth provider, the route gate and the
notification provider had none.

Two rules were worse than uncovered — they were *apparently* covered. BR-6 (one progress entry per
upgrade per date) and BR-14 (concurrent edits are refused) are enforced by a unique index and a
`@Version` column, and both were "tested" by service tests that mock the repository. Those tests pass
just as happily against a table with no constraint at all.

## The approach

Four test levels with a rule for choosing between them — write at the cheapest level that can actually
observe the behaviour, and never at one that cannot — recorded in
[ADR-009](docs/ADRs/ADR-009-test-levels-boundaries-and-naming.md). Mock our own ports; never mock
PostgreSQL. Slice tests run the **real** security chain, because a slice with security disabled asserts
the opposite of FR-5 while looking green.

The integration tests now start their own database
([ADR-008](docs/ADRs/ADR-008-testcontainers-for-the-integration-test-database.md)). Previously `mvn
verify` read `DB_*` and defaulted to `localhost:5432`, so anything already listening was accepted
silently — on this machine a native PostgreSQL 16 service owns that port, so `docker-compose up -d
postgres` lost the binding and the suite ran against the developer's own database. That had to be fixed
before the BR-6 and BR-14 tests were worth writing.

## What was done

| | Before | After |
|---|---|---|
| Backend unit + slice tests | 175 | 423 |
| Backend integration tests | 5 | 20 |
| Frontend tests | 61 | 108 |
| Backend line coverage | 43% | 87% |
| Backend branch coverage | 65% | 83% |
| Requirements naming a real test | a minority | 70 of 77 |

The seven that do not: four name the command, workflow or script that *is* the check (NFR-11, NFR-12,
NFR-13, NFR-18); three — FR-39, NFR-19, NFR-20 — are hand-verified and say so, because each is about a
rendered width, a colour or an overflow, and jsdom has no layout engine to observe any of them.

The existing backend suite was also renamed onto `Given…_When…_Then…` in one rename-only commit — the
frontend already used it, so one of the two conventions had to move.

## Two bugs the tests found

Both are fixed in their own `fix:` commits, each with the test that found it.

**`GET /api/auth/me` required no token.** It sat behind a blanket
`.requestMatchers("/api/auth/**").permitAll()`, so an anonymous request reached the controller, whose
`@AuthenticationPrincipal` was null, and the dereference answered **500**. FR-5 says every endpoint but
registration, login and the health checks requires a token; this one did not. The rule now names the two
public endpoints exactly.

**A wrong password answered 500.** `AuthController` authenticates inside a handler method rather than in
the security chain, so `BadCredentialsException` reached `GlobalExceptionHandler`, which mapped no
`AuthenticationException` and fell through to the catch-all. The most ordinary outcome the login endpoint
has was reported as a server fault and logged with a full stack trace. `AuthController`'s own javadoc
already documented it as a 401; it now is one, with a generic message so the response does not say which
half was wrong.

One documentation error was corrected the same way: `HealthAreaService` said upgrades filed under a
deleted area "keep the now-dangling area id", but the schema says `area_id ... ON DELETE SET NULL`, so
they survive and become unfiled. `HealthAreaPersistenceIT` now holds that.

## Trade-offs accepted

- **`mvn verify` now requires a running Docker daemon.** A hard requirement with a clear error, in place
  of a silent wrong answer. `mvn test` is untouched and still needs neither a database nor Docker
  (NFR-11).
- **Testcontainers is pinned to 1.21.4**, overriding the Boot BOM's 1.19.7, whose docker-java negotiates
  Docker API 1.32 — Docker Engine 29 refuses anything below 1.40 and the container never starts.
- **Coverage is reported and gates nothing.** The gate is `requirements.md` naming a test per
  requirement, which is a human check. A threshold is satisfied by tests written to move a number.
- **No real-browser gate.** FR-39, NFR-19 and NFR-20 stay hand-verified and stay in §6; Playwright drags
  the backend, a database and a browser binary into CI and deserves its own decision.

## Two things left open, deliberately

Both are recorded in requirements.md §6 rather than changed here, because each is a decision rather than
a fix:

1. **An anonymous request to a protected endpoint answers 403, not 401** — no `AuthenticationEntryPoint`
   is configured. FR-5 is met either way, but a client cannot distinguish "not signed in" from "not
   allowed", and changing it breaks anyone branching on the status.
2. **The daily check-in posts an entry for every active upgrade, including untouched ones** — which under
   BR-6 records them as a failure that cannot be corrected until the next day. `DailyCheckinPage`'s own
   comment says the opposite. FR-19 does not settle which behaviour is right.

## How it was verified

Everything CI runs, locally, from a clean tree:

```
cd backend  && mvn -B clean verify                # 423 unit + 20 integration, green
cd frontend && npm ci && npm run lint && npm run check:colours
cd frontend && npm run test:coverage && npm run build
python3 docs/architecture/arch-diagrams/generate.py --check
```

Plus the checks that matter more than a green build:

- Every test class named in `requirements.md` was verified to exist (51 distinct names across 70 rows, none missing).
- NFR-11 still holds: only the five `*IT` classes extend the Testcontainers base and Failsafe alone runs
  those, so nothing in `mvn test` can reach a container. It was also run green with the daemon stopped
  while this branch was being built.
- The JaCoCo report was read: exactly one package sits at zero line coverage
  (`upgrade.application.port.out`, a single-line record). Reading it also turned up that the report
  counted only the unit JVM — the integration run is now merged in, so the persistence adapters no
  longer read as untested when they are not.

## Records

- [ADR-008](docs/ADRs/ADR-008-testcontainers-for-the-integration-test-database.md) — why the integration
  tests start their own database, and why not H2.
- [ADR-009](docs/ADRs/ADR-009-test-levels-boundaries-and-naming.md) — the four levels, their boundaries,
  the naming convention, and why coverage is reported rather than gated.


NaimElijah (commented):
## Review

Twelve findings. One is a regression this PR introduces, one is a bug it documents but does not track,
four are tests or docs that claim more than they deliver, and the rest are contract documentation left
behind by the code change.

---

### 1. `AuthenticationException → 401` swallows genuine server faults — and logs nothing

`GlobalExceptionHandler.java:143`

`InternalAuthenticationServiceException extends AuthenticationServiceException extends
AuthenticationException` (verified against spring-security-core). So when Postgres is down and
`UserDetailsServiceImpl` throws, `DaoAuthenticationProvider` wraps it and this handler answers
**401 "Invalid credentials"**.

Before this PR that fell through to `handleGeneral` — a 500 with `log.error` and the stack trace. Now
the user is told their password is wrong, the `ex` parameter is never read, and the outage leaves no
trace anywhere.

This contradicts NFR-7 ("Only a genuine server fault is a 500") and the standing rule "never swallow an
error". Narrow the handler to `BadCredentialsException` / `AccountStatusException` and let
`AuthenticationServiceException` fall through.

### 2. A deleted account answers 500, and the note it points at was never written

`JwtAuthenticationFilterTest.java:93`

The test comment says "Asserted as it behaves, not as it should behave; see the note on this in the pull
request". No such note exists — §6 and the PR body record only the 403-vs-401 and daily-check-in items.

Meanwhile `requirements.md` names this very test as NFR-5's enforcement, so the suite asserts a
violation of NFR-7 as correct with no tracked owner. The filter's own javadoc says "a missing or invalid
token is not rejected here: the request simply stays anonymous" — a token for a deleted user *is* an
invalid token, so catching `UsernameNotFoundException` and continuing anonymously is what the filter
already claims to do.

### 3. The FR-5 route guard compares paths, not methods

`AuthenticatedBoundaryTest.java:158`

`getPatternValues()` alone means adding `@DeleteMapping("/{id}")` to an already-listed path leaves the
pattern set unchanged and the guard green — while the class doc claims "a controller method added
without a row here means an endpoint whose authenticated boundary nothing checks, and this is what says
so". Build the set from `getMethodsCondition()` × `getPatternValues()`.

### 4. "Five most recent" does not test the sort

`DashboardAggregationServiceTest.java:184`

The fixture is already newest-first, so deleting `.sorted(...)` from
`DashboardAggregationService:61` leaves this green. FR-27 would then show the five *oldest* completed
upgrades with nothing red.

### 5. The difficulty fallback test asserts an expression written in the test

`upgradeMeta.test.ts:28`

`difficultyVariant[fromApi] ?? UNKNOWN_DIFFICULTY_VARIANT` only proves the constant is `'gray'`. The
`??` lives at two call sites; delete it from either and the colourless-badge regression the module was
created to prevent returns, with this test still passing. The fallback belongs in `upgradeMeta.ts`.

### 6. `report-merged` never runs when an integration test fails

`pom.xml:253`

Bound to `verify`, the same phase as `failsafe:verify`, which is declared first. One failing `*IT` and
the merge never happens — leaving the unit-only report on disk, which the `if: always()` summary step
then publishes under "Backend coverage" with the persistence adapters near 0%. That is exactly the
"report that lies in the direction of write more tests for something already tested" the pom comment
says this configuration exists to prevent.

### 7. The coverage summary crashes instead of degrading

`ci.yml:47`

`if: always()` + `mvn clean verify` + a unit-test failure = no `target/` and no CSV. The unguarded
`open()` raises `FileNotFoundError`, so the job shows a second red step unrelated to the real failure.

### 8–11. Documentation the code change left behind

- **`JwtAuthenticationFilter.java:36`** still names `/api/auth/**` as permitted and says everything else
  "returns 401". This PR narrowed that to two endpoints, and its own tests assert 403. `/ws/**` is
  missing too.
- **`GlobalExceptionHandler.java:39`** — the class table the javadoc calls the whole exception→status
  contract has no row for the new 401. A maintainer reading it concludes there is no 401 in the API and
  adds a second handler, which is the ambiguous mapping the paragraph below the table warns about.
- **`requirements.md:149`** — NFR-7 enumerates the wire statuses and does not mention the 401 this PR
  introduces. A new response status is a change to the contract.
- **`HealthAreaServiceTest.java:32`** repeats the "upgrades keep the now-dangling area id" claim that
  this same PR corrects in `HealthAreaService`, and that `HealthAreaPersistenceIT` asserts is false.

### 12. Nine redundant `@Import(GlobalExceptionHandler.class)`

`@WebMvcTest` component-scans `@RestControllerAdvice` already — confirmed by `AuthenticatedBoundaryTest`,
which does not import it and still gets the mappings. The import registers a second identical advice per
slice. Harmless, but it is nine copies of a line that says something untrue about how the slice is wired.

---

**Verdict:** findings 1 and 2 should block. The rest are worth fixing in the same pass — several are
places where a test or a document claims more than it delivers, which is the specific failure mode this
PR exists to remove.

NaimElijah:
## All twelve findings addressed

Six commits, `1388d30`..`3ee5c54`.

**The two blockers**

- **`3ee5c54`** — `AuthenticationException` narrowed to `BadCredentialsException` /
  `AccountStatusException`. Anything else authentication-shaped (a database outage during login arrives
  as `InternalAuthenticationServiceException`) falls through to `handleGeneral`: 500, logged, stack
  trace. The 401 itself still logs nothing on purpose — a wrong password is expected on a public
  endpoint and the submitted email is personal data (NFR-6). Added the contract-table row and NFR-7.
- **`1426861`** — the filter now catches `UsernameNotFoundException` and leaves the request anonymous,
  which is what its own contract already said should happen to an unusable token. The test asserts the
  requirement instead of the 500, so nothing needed adding to §6.

**The rest**

- **`3560c9d`** — route guard compares `METHOD + path`; the dashboard five-most-recent case is now fed
  oldest-first and checks the order; `HealthAreaServiceTest`'s javadoc corrected.
  The sort fix was verified by mutation: deleting `.sorted(...)` from
  `DashboardAggregationService` now fails the test and passed before.
- **`a685289`** — the difficulty fallback moved into `difficultyBadgeVariant`, used by both call sites;
  the test calls production code rather than an expression it wrote itself. It also covers a difficulty
  that is simply absent, which is legal and reached neither `??`.
- **`ed8ec1d`** — merge and report moved to `post-integration-test` so a failing IT can no longer leave
  the unit-only report to be published as merged; the summary step degrades instead of crashing when
  there is no CSV.
- **`1388d30`** — the nine redundant `@Import(GlobalExceptionHandler.class)` removed.

**Verified:** `mvn clean verify` — 425 unit + 20 integration, green, merged coverage 87% line / 83%
branch. Frontend — 109 tests, lint, colour check and build all clean.

The two items in §6 are unchanged and still open by choice: the 403-vs-401 entry point, and whether the
daily check-in should post entries for untouched upgrades.

PR #50, opened 2026-08-25
obs: give the system a log worth reading, an audit trail, metrics and probes

### LINKED ISSUE #49: obs: nothing is logged, measured or audited, so a running system cannot be explained
Split out of nothing — this is the gap [ADR-007](https://github.com/NaimElijah/UpHealther/blob/dev/docs/ADRs/ADR-007-request-correlation-through-micrometer-tracing.md)
left behind. Correlation was built so a failure could be *found*; almost nothing is written for it to
find, nothing is measured, and nothing records who did what.

## Verified

- `grep -rn 'log\.\(trace\|debug\|info\|warn\|error\)' backend/src/main` returns **5 lines, in 2
  classes** — `GlobalExceptionHandler` (4) and `SpringDomainEventPublisher` (1). Every controller,
  every application service, both schedulers, all four security classes and every persistence adapter
  are silent.
- `grep -rn 'EnableJpaAuditing\|CreatedBy\|Audited' backend/src` returns nothing, and no migration in
  `V1`–`V5` creates an audit table. The seven entities carry `createdAt`/`updatedAt` written by
  `@PrePersist`, which is row state, not an audit trail.
- `application.yml` has no `management.endpoints.web.exposure.include`, so only `/actuator/health` is
  reachable; `management.endpoint.health.probes.enabled` is unset, so `/actuator/health/liveness` and
  `/actuator/health/readiness` do not exist. There is no metrics registry on the classpath.
- Neither `backend/Dockerfile` nor `frontend/Dockerfile` has a `HEALTHCHECK`, and the `frontend`
  compose service has none either and does not wait for the backend to be healthy.
- `logging.level.com.healthupgrades: DEBUG` is the only logging configuration, and there are no
  profile files, so it is the production setting too.
- `grep -rn 'console\.\|ErrorBoundary\|x-trace-id\|traceId' frontend/src` returns nothing. Seven pages
  render a hard-coded `Failed to load X.` and discard the error object.

## Impact

**A running system cannot be explained after the fact.** A user says "it broke". The trace id in their
screenshot resolves to at most one ERROR line and its stack trace — there is no record of what the
request was trying to do, what state the aggregate was in, whether the transition was refused, or how
long anything took.

**Security events are invisible.** `AuthService` never records a successful login.
`JwtTokenProvider:86` swallows a forged or expired token silently. `JwtAuthenticationFilter:67`
silently leaves a token naming a deleted account anonymous. Brute force, credential stuffing and token
replay against deleted accounts leave no trace at all, so there is nothing to alert on and nothing to
review afterwards.

**Nothing is measured.** No latency, no error rate, no saturation, no connection-pool depth, no
domain success signal. `spring-boot-starter-actuator` is already a direct dependency and none of what
it collects is reachable.

**The frontend throws away the one thing that makes a report actionable.** The backend sets
`X-Trace-Id` on every response and stamps `traceId` on every error body (NFR-21). `src/api/client.ts`
reads neither, and no page displays either, so a user has nothing to quote. A render-time throw blanks
the page silently, because there is no error boundary anywhere in the tree.

## Sketch

- A `logback-spring.xml` that keeps Boot's stock pattern locally and switches to a JSON encoder under
  a profile — **including `defaults.xml`**, or ADR-007's `${LOG_CORRELATION_PATTERN}` is silently lost.
- A log-level policy where each level means one thing, and log statements written to it: scheduled-job
  outcomes at INFO, degraded delivery at WARN, a job that threw at ERROR.
- An `AuditTrail` outbound port in `common/domain/port/out/`, following the
  `DomainEventPublisher` precedent, written as a dedicated structured log stream. Called from the
  application services so that **refusals** are recorded too, which a domain-event listener could
  never see. Actor, action, resource, outcome, trace id — and no free-text field, so NFR-6 is enforced
  by the type rather than by a reviewer.
- `micrometer-registry-prometheus`, an explicit and closed actuator exposure list, and liveness /
  readiness probes. `HEALTHCHECK` in both images.
- The frontend reading `traceId` off the error body and the `X-Trace-Id` header, showing it beside the
  failure, and an error boundary so a throw is a message rather than a blank page.

## Acceptance criteria

- [ ] Log output is JSON in a container and readable locally, and a plain-text line still carries its
      `[traceId-spanId]` prefix — regression-guarded by a test, not by inspection
- [ ] Every level has one meaning, and the statements written match it
- [ ] Every state-changing use case and every authentication outcome records an audit event carrying
      actor, action, resource, outcome and trace id
- [ ] An audit event cannot carry personal data, because the type has nowhere to put it
- [ ] A refused operation is audited, not only a successful one
- [ ] `/actuator/health/liveness` and `/actuator/health/readiness` answer, and `/actuator/env` and
      `/actuator/heapdump` do not
- [ ] `/actuator/prometheus` exposes latency, error rate, saturation, pool depth and a domain signal,
      with no unbounded tag
- [ ] Both images declare a `HEALTHCHECK`, and the frontend waits for a healthy backend
- [ ] A user who sees a failure in the UI is shown a trace id that finds the log line
- [ ] A render-time throw shows a recoverable message instead of a blank page
- [ ] `requirements.md`, `architecture.md` and the ADRs record the contract in the same commit as the
      behaviour

NaimElijah:
Closes #49.

## The problem

[ADR-007](https://github.com/NaimElijah/UpHealther/blob/dev/docs/ADRs/ADR-007-request-correlation-through-micrometer-tracing.md)
built correlation and recorded, in passing, that there were **five log statements in `src/main`**. That
number never moved. So the trace ids worked and there was almost nothing to correlate: no controller,
no application service, neither scheduler, no security class and no persistence adapter wrote a line.

Around that gap were four more. Nothing recorded **who did what** — no JPA auditing, no Envers, no audit
table, and BR-15 answers an attempt on somebody else's record with a `404`, so reaching into another
user's data left no trace anywhere. Actuator had been a direct dependency since ADR-007 and everything
it collects was **unreachable**, with no readiness or liveness probe and no `HEALTHCHECK` in either
image. `logging.level.com.healthupgrades: DEBUG` was the only logging configuration and therefore the
production one. And the frontend **discarded the trace id** the backend goes out of its way to produce:
seven pages rendered a hard-coded `Failed to load X.` and threw the response away.

## The approach

Four decisions, each with its own ADR, and each recording what would reverse it.

| | Decision | Record |
|---|---|---|
| **Logs** | `logback-spring.xml` picks a JSON encoder under the `json-logs` profile and Boot's readable pattern everywhere else; each level means one thing; nothing personal in a line | [ADR-010](https://github.com/NaimElijah/UpHealther/blob/feat/observability-logging-and-audit/docs/ADRs/ADR-010-structured-logging-and-a-level-policy.md) |
| **Audit** | An `AuditTrail` outbound port written as a log stream, recording refusals as well as successes | [ADR-011](https://github.com/NaimElijah/UpHealther/blob/feat/observability-logging-and-audit/docs/ADRs/ADR-011-audit-as-a-log-stream.md) |
| **Metrics** | `/actuator/prometheus`, liveness and readiness, and an actuator surface closed by name | [ADR-012](https://github.com/NaimElijah/UpHealther/blob/feat/observability-logging-and-audit/docs/ADRs/ADR-012-metrics-through-a-prometheus-scrape-endpoint.md) |
| **The browser** | The trace id reaches the screen, and a render error is a message rather than a blank page | NFR-30, NFR-31 |

Two choices are load-bearing and easy to get wrong later:

- **`logback-spring.xml` imports Boot's `defaults.xml` rather than writing a `<pattern>`.** That import
  is what carries `${LOG_CORRELATION_PATTERN}` into the output. A hand-written pattern compiles, starts,
  logs happily and correlates nothing — the exact regression ADR-007 warned about. `LogOutputFormatTest`
  renders one event through the encoder the shipped file installs, in *both* formats, so the mistake
  fails a build.
- **`AuditEvent` has nowhere to put free text.** Two enums and two identifiers. NFR-6 becomes a property
  of the type rather than of whoever reviews the next call site, and `AuditEventTest` asserts it by
  inspecting the record's components — so a future `String` field fails at the moment it is added.

## Trade-offs accepted

- **A log stream, not an `audit_log` table.** An entry has to outlive the transaction it observes: a
  refused transition rolls back, and a row written inside it would roll back too, losing exactly the
  entry somebody was looking for. The cost is that the trail's retention is `docker logs`' retention,
  which is adequate for diagnosis and explicitly **not** a compliance story. Revisit trigger is in
  ADR-011.
- **A refused login is recorded with no subject at all.** The submitted email is personal data and
  naming a user id would confirm the account exists. That leaves a rate signal rather than an
  attribution, and it will not support a lockout policy as written — stated as a limit rather than
  glossed.
- **`/actuator/**` stays `permitAll`.** The textbook answer is a separate management port; ADR-012
  records why that was judged premature and names the trigger, which it shares with ADR-007's
  trusted-proxy boundary. With the exposure list closed by name and asserted by test, what remains
  readable is a status, build info and counters.
- **Nothing scrapes the metrics endpoint and no trace leaves the process.** Both are deliberate stopping
  points. The missing piece in each case is a deployment somebody is paged for.
- **The browser can show an error and not record it.** No telemetry, no `console`, no reporting
  endpoint — adding one is a decision about sending user data off the device.

## One deliberate behaviour change

A STOMP push that fails is now caught, reported at WARN and not rethrown. It runs from an `afterCommit`
callback, so the notification is already durable and FR-32's "readable afterwards regardless" is already
met by the row. Letting the exception out failed the caller *after* its transaction had committed, and
in `dispatchReminders` that meant **one unreachable session cancelled everybody else's reminders for
that minute**. Nothing recorded it either way.

## What was done

- **Logging** — `logback-spring.xml` (profile-switched, `logstash-logback-encoder` pinned to 7.4 against
  logback 1.4.14, verified from the dependency tree); default level `INFO`, `LOG_LEVEL_APP` to raise it;
  scheduler summaries, security-boundary lines and the push failure, all written to the level policy.
- **Audit** — `AuditTrail` port, `AuditAction`/`AuditedResource`/`AuditOutcome`/`AuditEvent`,
  `LoggingAuditTrail` (which derives `audit.events{action,outcome}` from the same call), and a recording
  call in 21 use cases across seven services plus both auth outcomes.
- **Metrics and health** — `micrometer-registry-prometheus`; exposure closed to `health,info,prometheus`;
  liveness and readiness enabled; `JobMetrics` timing and counting all three scheduled jobs;
  `HEALTHCHECK` in both images; the frontend now waits for `service_healthy` rather than for the
  container to exist.
- **Frontend** — `api/apiError.ts` decodes the error contract once (body first, `X-Trace-Id` header as
  fallback, because a security-chain rejection carries only the header); `ui/ErrorState` puts the id on
  the screen across seven pages; `ErrorBoundary`; `RegisterPage` stops answering "email may already be
  in use" to a 500.
- **Architecture fitness** — ArchUnit rule 1 gains `org.slf4j` and `io.micrometer`; neither was caught by
  any existing entry, so a domain class could have picked up a logger unnoticed.
- **Docs** — three ADRs; ADR-007 marked superseded in part (it claimed there would be no
  `logback-spring.xml`); NFR-22…NFR-31 with their enforcing tests; a top-level `## Observability` section
  in `architecture.md` replacing the `### Correlation` subsection; four new entries in Known constraints;
  README observability section and the env-var table completed; both `CLAUDE.md` files.

## How it was verified

**Automated**

```
mvn -B clean verify          # 461 unit + 32 integration, green
npm run lint && npm run check:colours && npx vitest run && npm run build   # 125 tests, green
python docs/architecture/arch-diagrams/generate.py --check                 # no drift
```

New coverage: `LogOutputFormatTest` (both formats through the real encoder), `AuditEventTest`,
`AuditOutcomeTest`, `AuditTrailTest`, `LoggingAuditTrailTest`, `JobMetricsTest`, `ActuatorEndpointsIT`
(including seven "not exposed" assertions), `apiError.test.ts`, `ErrorState.test.tsx`,
`ErrorBoundary.test.tsx`.

**End to end, against `docker-compose up --build`**

- Compose brought the stack up `db healthy → backend healthy → frontend`, which is the readiness gate
  doing its job.
- A wrong password returned `{"status":401,...,"traceId":"ab43d9b8…"}`, and that exact id was on the
  audit line `action=auth.login outcome=REFUSED actorId=none resourceId=none` — **no email anywhere**.
- Creating an upgrade and then illegally pausing it returned `traceId: fb6ba358…`; grepping the log for
  it found `action=upgrade.pause outcome=REFUSED actorId=a0000000-… resourceId=7d57ce1b-…`. That round
  trip — a user quotes one string, and it names who tried what and that it was refused — is the whole
  point of the change.
- `/actuator/health/{liveness,readiness}` both `UP`; `env`, `heapdump`, `loggers`, `beans`,
  `configprops`, `threaddump` and `mappings` all `404`.
- `/actuator/prometheus` served `audit_events_total{action,outcome}`,
  `scheduled_job_runs_total{job,outcome}`, HTTP, JVM and HikariCP metrics, all tagged
  `application="health-upgrades-tracker"`.
- Grepped the whole container log: **0** lines containing the demo email, **0** containing an upgrade
  title. The one `password` match is the Flyway migration *name* `V3__fix_demo_password`.
- Reading that log also found the last commit's bug: Boot prints its banner to stdout before logback
  exists, so nine of the first sixty lines were unparseable JSON. Fixed; the log is now **38 lines, 0
  non-JSON**.

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -29,6 +29,12 @@
           a major release and spring-boot-testcontainers 3.2.5 is built against 1.x. See ADR-008.
         -->
         <testcontainers.version>1.21.4</testcontainers.version>
+        <!--
+          Not managed by spring-boot-dependencies, so the version is ours to keep correct. 7.4 is the
+          last release built against logback 1.3/1.4; 8.x compiles against logback 1.5, which Boot 3.2.5
+          does not ship (it resolves logback 1.4.14). Re-check this pin on any Boot upgrade. See ADR-010.
+        -->
+        <logstash.encoder.version>7.4</logstash.encoder.version>
     </properties>
 
     <dependencies>
@@ -109,6 +115,17 @@
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-tracing-bridge-otel</artifactId>
+        </dependency>
+
+        <!--
+          The JSON encoder logback-spring.xml selects under the json-logs profile. A container's log is
+          read by a machine before a person, and the trace id, the level and the audit fields are only
+          queryable if they are fields rather than substrings of a message. See ADR-010.
+        -->
+        <dependency>
+            <groupId>net.logstash.logback</groupId>
+            <artifactId>logstash-logback-encoder</artifactId>
+            <version>${logstash.encoder.version}</version>
         </dependency>
 
         <!-- Test -->
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -128,6 +128,17 @@
             <version>${logstash.encoder.version}</version>
         </dependency>
 
+        <!--
+          Turns what actuator already collects - request latency, error rate, JVM and connection-pool
+          saturation - into something a scraper can read, alongside this application's own counters.
+          A pull endpoint, so it needs no collector to exist before it is useful. Version managed by
+          spring-boot-dependencies. See ADR-012.
+        -->
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+        </dependency>
+
         <!-- Test -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/backend/src/main/java/com/healthupgrades/auth/application/AuthService.java
+++ b/backend/src/main/java/com/healthupgrades/auth/application/AuthService.java
@@ -1,12 +1,18 @@
 package com.healthupgrades.auth.application;
 
+import com.healthupgrades.common.domain.audit.AuditAction;
+import com.healthupgrades.common.domain.audit.AuditEvent;
+import com.healthupgrades.common.domain.audit.AuditOutcome;
 import com.healthupgrades.common.domain.exception.BusinessRuleException;
+import com.healthupgrades.common.domain.port.out.AuditTrail;
 import com.healthupgrades.common.security.JwtTokenProvider;
 import com.healthupgrades.user.application.port.in.UserCommand;
 import com.healthupgrades.user.application.port.in.UserQuery;
 import com.healthupgrades.user.domain.model.User;
 import lombok.RequiredArgsConstructor;
+import org.springframework.security.authentication.AccountStatusException;
 import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
@@ -27,6 +33,7 @@ public class AuthService {
     private final PasswordEncoder passwordEncoder; // BCrypt encoder
     private final JwtTokenProvider tokenProvider; // issues JWTs
     private final AuthenticationManager authenticationManager; // verifies credentials on login
+    private final AuditTrail auditTrail; // records who signed in, and who was turned away
 
     /**
      * Registers a new user and issues a token for them.
@@ -41,16 +48,25 @@ public class AuthService {
      */
     @Transactional
     public AuthResult register(String name, String email, String password) {
-        if (userQuery.existsByEmail(email)) {
-            throw new BusinessRuleException("Email already registered: " + email);
+        // Recorded by hand rather than through AuditTrail.recording, because until the save returns
+        // there is no user id to name as the actor, and the email that would identify the attempt is
+        // exactly what NFR-6 says must not be written down.
+        try {
+            if (userQuery.existsByEmail(email)) {
+                throw new BusinessRuleException("Email already registered: " + email);
+            }
+            User user = User.builder()
+                    .name(name)
+                    .email(email)
+                    .passwordHash(passwordEncoder.encode(password)) // never store the raw password
+                    .build();
+            user = userCommand.save(user);
+            auditTrail.record(AuditEvent.allowed(AuditAction.AUTH_REGISTER, user.getId(), user.getId()));
+            return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
+        } catch (RuntimeException thrown) {
+            auditTrail.record(AuditEvent.from(AuditAction.AUTH_REGISTER, null, null, thrown));
+            throw thrown;
         }
-        User user = User.builder()
-                .name(name)
-                .email(email)
-                .passwordHash(passwordEncoder.encode(password)) // never store the raw password
-                .build();
-        user = userCommand.save(user);
-        return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
     }
 
     /**
@@ -64,10 +80,25 @@ public class AuthService {
      *         only reachable if the account is deleted mid-request
      */
     public AuthResult login(String email, String password) {
-        authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(email, password)); // throws on bad creds
-        User user = userQuery.findByEmail(email)
-                .orElseThrow(() -> new BusinessRuleException("User not found"));
-        return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
+        try {
+            authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(email, password)); // throws on bad creds
+            User user = userQuery.findByEmail(email)
+                    .orElseThrow(() -> new BusinessRuleException("User not found"));
+            auditTrail.record(AuditEvent.allowed(AuditAction.AUTH_LOGIN, user.getId(), user.getId()));
+            return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
+        } catch (BadCredentialsException | AccountStatusException refused) {
+            // A refused login is recorded with no subject at all. The submitted email is personal data
+            // (NFR-6) and there is no user id to name, because naming one would mean confirming that
+            // the account exists. What survives is the count and the trace id — a rate signal, not an
+            // attribution, which is the honest limit of auditing an anonymous endpoint.
+            auditTrail.record(new AuditEvent(AuditAction.AUTH_LOGIN, null, null, AuditOutcome.REFUSED));
+            throw refused;
+        } catch (RuntimeException thrown) {
+            // Everything else is a fault, not a rejection - the same distinction GlobalExceptionHandler
+            // draws when it refuses to report a database outage to the user as a wrong password.
+            auditTrail.record(AuditEvent.from(AuditAction.AUTH_LOGIN, null, null, thrown));
+            throw thrown;
+        }
     }
 
     /**
--- a/backend/src/main/java/com/healthupgrades/auth/application/AuthService.java
+++ b/backend/src/main/java/com/healthupgrades/auth/application/AuthService.java
@@ -61,8 +61,12 @@ public class AuthService {
                     .passwordHash(passwordEncoder.encode(password)) // never store the raw password
                     .build();
             user = userCommand.save(user);
+            // Recorded after the token exists, not before. Issuing it can fail - a secret too short to
+            // sign with - and recording ALLOWED first would then put one attempt in the trail twice,
+            // once as allowed and once as failed, with the counter double-counting to match.
+            AuthResult result = new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
             auditTrail.record(AuditEvent.allowed(AuditAction.AUTH_REGISTER, user.getId(), user.getId()));
-            return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
+            return result;
         } catch (RuntimeException thrown) {
             auditTrail.record(AuditEvent.from(AuditAction.AUTH_REGISTER, null, null, thrown));
             throw thrown;
@@ -84,8 +88,11 @@ public class AuthService {
             authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(email, password)); // throws on bad creds
             User user = userQuery.findByEmail(email)
                     .orElseThrow(() -> new BusinessRuleException("User not found"));
+            // As in register: after the token, so a signing failure cannot produce two entries for one
+            // sign-in attempt.
+            AuthResult result = new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
             auditTrail.record(AuditEvent.allowed(AuditAction.AUTH_LOGIN, user.getId(), user.getId()));
-            return new AuthResult(tokenProvider.generateToken(user.getEmail()), user);
+            return result;
         } catch (BadCredentialsException | AccountStatusException refused) {
             // A refused login is recorded with no subject at all. The submitted email is personal data
             // (NFR-6) and there is no user id to name, because naming one would mean confirming that
--- a/backend/src/main/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrail.java
+++ b/backend/src/main/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrail.java
@@ -1,12 +1,15 @@
 package com.healthupgrades.common.adapter.out.audit;
 
 import com.healthupgrades.common.domain.audit.AuditEvent;
+import com.healthupgrades.common.domain.audit.AuditOutcome;
 import com.healthupgrades.common.domain.port.out.AuditTrail;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 import java.util.UUID;
 
@@ -47,30 +50,73 @@ public class LoggingAuditTrail implements AuditTrail {
     /**
      * {@inheritDoc}
      *
-     * <p>INFO, because an audit entry is a state transition rather than a fault — a refusal is the
-     * system working. The level policy is in
-     * {@code docs/ADRs/ADR-010-structured-logging-and-a-level-policy.md}.
+     * <p><b>An {@code ALLOWED} entry waits for the commit.</b> Every caller wraps the body of an
+     * {@code @Transactional} service method, so the work returns while its transaction is still open —
+     * the commit happens afterwards, in the proxy. No repository adapter flushes, so a {@code @Version}
+     * clash or a unique constraint losing a race is decided <em>at commit</em>, after the service method
+     * has returned. Writing the entry at that point would have the trail and the counter both claim an
+     * edit succeeded while the caller was answered 409 and nothing was persisted. Deferring it through
+     * the same {@code afterCommit} route {@code NotificationService} already uses is what makes
+     * "allowed" mean what {@code ADR-011} says it means.
      *
-     * <p>The counter is derived here rather than incremented separately at the call site so the two can
-     * never disagree about what happened. Its tags are the action key and the outcome, both closed
-     * enums: the actor and the resource are deliberately not tags, because a user id as a label is an
-     * unbounded cardinality and a metrics backend is the wrong place to look one up anyway.
+     * <p>Refusals and faults are written immediately and deliberately: they already describe an attempt
+     * that did not land, they are the security-relevant half, and a process that dies before commit
+     * should not take them with it.
      */
     @Override
     public void record(AuditEvent event) {
-        log.info("{} {} {} {} {}",
-                keyValue("action", event.action().key()),
-                keyValue("outcome", event.outcome()),
-                keyValue("resource", event.action().resource()),
-                keyValue("actorId", idOrNone(event.actorUserId())),
-                keyValue("resourceId", idOrNone(event.resourceId())));
+        if (event.outcome() == AuditOutcome.ALLOWED && TransactionSynchronizationManager.isSynchronizationActive()) {
+            TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+                @Override
+                public void afterCompletion(int status) {
+                    // afterCompletion rather than afterCommit, so a rollback is recorded rather than
+                    // leaving the attempt with no entry at all.
+                    write(status == STATUS_COMMITTED ? event : event.refused());
+                }
+            });
+            return;
+        }
+        write(event);
+    }
 
-        Counter.builder(COUNTER)
-                .description("Audited attempts, by what was attempted and how it ended")
-                .tag("action", event.action().key())
-                .tag("outcome", event.outcome().name())
-                .register(meterRegistry)
-                .increment();
+    /**
+     * Writes one entry, and never lets a failure to do so fail the work being audited.
+     *
+     * <p>The port says implementations must not throw, and this is where that is honoured. It matters
+     * most on the refusal path: {@code AuditTrail.recording} calls this from inside its catch block, so
+     * an exception escaping here would <em>replace</em> the business exception — turning a
+     * {@code BusinessRuleException} that should be a 422 into an unhandled 500 with the real cause lost.
+     *
+     * <p>The two halves are guarded separately so a meter failure still leaves the audit line, which is
+     * the half that matters. If the logging subsystem itself is what failed there is nowhere left to
+     * report it, and the second write will throw for the same reason the first did.
+     */
+    private void write(AuditEvent event) {
+        try {
+            log.info("{} {} {} {} {}",
+                    keyValue("action", event.action().key()),
+                    keyValue("outcome", event.outcome()),
+                    keyValue("resource", event.action().resource()),
+                    keyValue("actorId", idOrNone(event.actorUserId())),
+                    keyValue("resourceId", idOrNone(event.resourceId())));
+        } catch (RuntimeException unwritable) {
+            return; // Nothing can be reported about a broken logger, by a logger.
+        }
+
+        try {
+            // Derived from the same call as the line above, so the count and the trail cannot disagree.
+            // Tagged by action and outcome only: both are closed enums, whereas an actor or a resource
+            // id is an unbounded label, and a metrics backend is the wrong place to look one up anyway.
+            Counter.builder(COUNTER)
+                    .description("Audited attempts, by what was attempted and how it ended")
+                    .tag("action", event.action().key())
+                    .tag("outcome", event.outcome().name())
+                    .register(meterRegistry)
+                    .increment();
+        } catch (RuntimeException uncountable) {
+            log.error("{} was recorded but could not be counted",
+                    keyValue("action", event.action().key()), uncountable);
+        }
     }
 
     /**
--- a/backend/src/main/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrail.java
+++ b/backend/src/main/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrail.java
@@ -1,0 +1,85 @@
+package com.healthupgrades.common.adapter.out.audit;
+
+import com.healthupgrades.common.domain.audit.AuditEvent;
+import com.healthupgrades.common.domain.port.out.AuditTrail;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.util.UUID;
+
+import static net.logstash.logback.argument.StructuredArguments.keyValue;
+
+/**
+ * Adapter writing the audit trail to its own log stream, and counting the same events as it goes.
+ *
+ * <p><b>Why a log stream and not a table.</b> An audit entry has to survive the transaction it observes:
+ * a refused transition rolls back, and a row written inside that transaction would roll back with it —
+ * exactly the entry somebody was looking for. It also arrives already correlated, because the trace id
+ * is on the line, so an entry leads back to the request that produced it without a foreign key.
+ * {@code docs/ADRs/ADR-011-audit-as-a-log-stream.md} records the decision and what would reverse it.
+ *
+ * <p>The logger is named {@code AUDIT} rather than after this class so the stream can be routed or
+ * filtered on its own, by a level or by an appender, without knowing where in the package tree it is
+ * written from.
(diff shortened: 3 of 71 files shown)


backend/src/main/java/com/healthupgrades/common/domain/port/out/AuditTrail.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param action      what is being attempted
+     * @param actorUserId the user attempting it
+     * @param resourceId  the record acted on, or {@code null} when there is not one yet
+     * @param operation   the work to run and observe
+     * @return whatever the operation returned
+     */
+    default <T> T recording(AuditAction action, UUID actorUserId, UUID resourceId, Supplier<T> operation) {
+        try {
+            T result = operation.get();
+            record(AuditEvent.allowed(action, actorUserId, resourceId));
```
NaimElijah, line 42: **`ALLOWED` is recorded before the transaction commits, so the trail says "allowed" for work that never happened.**

Every caller wraps the *body* of an `@Transactional` service method, so `recording(...)` returns — and writes `ALLOWED` — while the transaction is still open. The commit happens afterwards, in the proxy, outside this try/catch.

Concrete: `UpgradeService.update` on a stale `@Version`. No adapter flushes (`UpgradeRepositoryAdapter.save` is a bare `jpa.save`), so the version check fires at commit. Sequence: lambda returns → `ALLOWED` logged and `audit.events{outcome="ALLOWED"}` incremented → commit throws `ObjectOptimisticLockingFailureException` → client gets 409 → nothing persisted. The audit trail and the metric both say the edit succeeded. Same for `AuthService.register` and `HealthAreaService.create` losing a unique-constraint race at commit, and for any `DataIntegrityViolationException`.

A knock-on: `AuditOutcome.of` lists `OptimisticLockException` as a refusal, but that exception can never reach this catch through `recording(...)` — `AuditOutcomeTest.GivenALostConcurrentEdit_...` is asserting an unreachable path.

Recording `ALLOWED` from an `afterCommit` synchronization (the pattern `NotificationService.pushAfterCommit` already uses) would make the entry mean what ADR-011 says it means. The `REFUSED`/`FAILED` half is correctly placed and should stay where it is.

backend/src/main/java/com/healthupgrades/common/observability/JobMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param job  the job's stable name, used as a metric tag
+     * @param work the job body
+     * @throws RuntimeException whatever the job threw, untouched, so the scheduler still reports it
+     */
+    public void timed(String job, Runnable work) {
+        Timer.Sample sample = Timer.start(meterRegistry);
+        String outcome = OK;
+        try {
+            work.run();
+        } catch (RuntimeException thrown) {
```
NaimElijah: **A job that dies on an `Error` is counted as `outcome="ok"`.**

`outcome` is initialised to `OK` and only reassigned in `catch (RuntimeException)`. An `Error` — `NoClassDefFoundError` from a missing transitive dep, `ExceptionInInitializerError`, `StackOverflowError`, `OutOfMemoryError` — skips the catch, runs the `finally`, and increments `scheduled.job.runs{outcome="ok"}` on its way out.

That is the exact failure this class exists to make visible: the sweep threw, nobody was answered with a 500, and the one signal says it ran fine. `scheduled.job.runs{outcome="failed"}` stays at zero forever while the job has not worked in a week.

Setting the outcome positively after the body returns closes it, and keeps the rethrow-untouched property:

```suggestion
        String outcome = FAILED;
        try {
            work.run();
            outcome = OK;
        } finally {
```

(Same class of gap, lower stakes, in `AuditTrail.recording`: an `Error` there leaves no audit entry at all.)

backend/src/main/resources/application.yml
```diff
@@ -49,6 +49,26 @@ app:
...
+      # Splits /actuator/health into /liveness (is the process alive) and /readiness (can it serve).
+      # A restart fixes the first and never the second, which is why an orchestrator needs both and
+      # why the compose health-check polls readiness rather than the aggregate.
+      probes:
+        enabled: true
+  observations:
+    key-values:
+      # Tags every metric and observation with the service, so two deployments are separable in one
+      # scrape target without parsing anything.
+      application: ${spring.application.name}
```
NaimElijah, line 79: **`management.observations.key-values` does not tag every metric — only observations.**

Boot 3.2.5's own configuration metadata draws the line explicitly:

- `management.observations.key-values` — *"Common key-values that are applied to every **observation**."*
- `management.metrics.tags` — *"Common tags that are applied to every **meter**."* (present and not deprecated in 3.2.5)

So `http_server_requests_seconds` gets `application="health-upgrades-tracker"`, and `jvm_memory_used_bytes`, `hikaricp_connections`, `audit.events` and `scheduled.job.runs` do **not** — they are plain meters, not observations. The comment above ("Tags every metric and observation"), README's scrape table and ADR-012's "two deployments separable in one scrape target" all overstate what this line buys.

`ActuatorEndpointsIT:93` does not catch it: `contains("application=…")` passes on the one HTTP timer that does carry the tag.

Adding `management.metrics.tags.application: ${spring.application.name}` alongside this covers the meters the observation property cannot reach.

backend/src/main/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrail.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void record(AuditEvent event) {
+        log.info("{} {} {} {} {}",
+                keyValue("action", event.action().key()),
+                keyValue("outcome", event.outcome()),
+                keyValue("resource", event.action().resource()),
+                keyValue("actorId", idOrNone(event.actorUserId())),
+                keyValue("resourceId", idOrNone(event.resourceId())));
+
+        Counter.builder(COUNTER)
```
NaimElijah: **`AuditTrail.record` is documented "Implementations must not throw: an audit failure must not fail the work" — this implementation can throw, and where it throws it destroys the original exception.**

`Counter.builder(...).register(meterRegistry)` throws (e.g. `IllegalArgumentException` when a meter of the same name already exists with a different type or tag set, or from a registry in a bad state), and `log.info` can throw from an appender.

The damage is not the audit line, it is where the throw lands. `AuditTrail.recording` calls `record(...)` **inside** its `catch (RuntimeException thrown)` block. If `record` throws there, the audit exception replaces `thrown` on its way up, and a `BusinessRuleException` that should have been a 422 becomes an unhandled 500 with the real cause gone.

Either honour the contract here (wrap the body so a failure to audit is swallowed or logged, never propagated), or drop the "must not throw" clause from the port and make `recording` guard its catch block. Right now the contract is stated and not held anywhere.

backend/src/main/java/com/healthupgrades/auth/application/AuthService.java
```diff
@@ -41,16 +48,25 @@ public class AuthService {
...
+            if (userQuery.existsByEmail(email)) {
+                throw new BusinessRuleException("Email already registered: " + email);
+            }
+            User user = User.builder()
+                    .name(name)
+                    .email(email)
+                    .passwordHash(passwordEncoder.encode(password)) // never store the raw password
+                    .build();
+            user = userCommand.save(user);
+            auditTrail.record(AuditEvent.allowed(AuditAction.AUTH_REGISTER, user.getId(), user.getId()));
```
NaimElijah, line 68: **One registration can produce two audit entries.**

`record(allowed)` runs, then `tokenProvider.generateToken(...)` runs — inside the same `try`. If token generation throws (a bad `JWT_SECRET`, a key-derivation failure), the catch fires and records a *second* entry, `AUTH_REGISTER/FAILED`, for the same attempt. `audit.events` then counts one registration twice, under two different outcomes, and `RecordingAuditTrail.only(AUTH_REGISTER)` throws its `AssertionError` if a test ever hits that path.

`login` (lines 87-88) has the identical shape.

Moving the `record(allowed)` call to after the `AuthResult` is built — or narrowing the `try` to the part being audited — makes one attempt produce one entry.

backend/src/test/java/com/healthupgrades/ActuatorEndpointsIT.java
```diff
@@ -0,0 +1,119 @@
...
+                .as("saturation: the JVM and the connection pool")
+                .contains("jvm_memory_used_bytes")
+                .contains("hikaricp_connections");
+    }
+
+    @Test
+    void GivenAScrape_WhenTheApplicationTagIsRead_ThenEveryMetricNamesThisService() throws Exception {
+        HttpResponse<String> response = get("/actuator/prometheus");
+
+        assertThat(response.body()).contains("application=\"health-upgrades-tracker\"");
```
NaimElijah: **The test is named `...ThenEveryMetricNamesThisService` but asserts that *one* substring appears somewhere in the scrape.**

`contains("application=…")` passes as soon as a single metric carries the tag — which is what happens today, because `management.observations.key-values` tags observations only (see the comment on `application.yml`). `jvm_memory_used_bytes`, `hikaricp_connections`, `audit.events` and `scheduled.job.runs` have no `application` label at all and this test is green.

A test that matched its name would pick a known non-observation meter and assert the tag on that line, e.g. that every `jvm_memory_used_bytes{...}` sample carries `application="health-upgrades-tracker"`. As written it certifies a property the configuration does not have.

frontend/src/api/apiError.test.ts
```diff
@@ -0,0 +1,87 @@
...
+ * stamps `traceId` on every error body it builds (NFR-21), so that one string finds the request in the
+ * log. Every call site used to discard it. The two sources are both covered because they have
+ * different reach: a request refused inside the security chain never reaches the exception handler
+ * that writes the body, so it comes back with the header alone.
+ */
+describe('toApiError', () => {
+  const config = { headers: new AxiosHeaders() } as InternalAxiosRequestConfig;
+
+  function responded(status: number, data: unknown, headers: Record<string, string> = {}): AxiosError {
+    const response = { data, status, statusText: '', headers, config } as AxiosResponse;
```
NaimElijah: **`headers` is stubbed as a plain object, so the header-fallback test never exercises the lookup it is guarding.**

At runtime axios hands `response.headers` an `AxiosHeaders` instance (keys normalised to lower case internally). Here it is a bare `Record<string, string>` cast through `as AxiosResponse`, so `GivenARejectionWithNoBody_...` proves only that `obj['x-trace-id']` reads back what the test just wrote under that exact key.

That is precisely the case the test exists for — a request refused inside the security chain, where the header is the *only* source of the trace id — and it would stay green if axios's real header object did not answer to the lower-cased bracket access `apiError.ts` uses.

`new AxiosHeaders({ 'X-Trace-Id': '…' })` (server casing, through the real type) tests the production path.

backend/src/main/java/com/healthupgrades/notification/adapter/in/scheduling/NotificationScheduler.java
```diff
@@ -88,27 +108,38 @@ public void notifyDailyCheckin() {
...
+                    fired++;
+                    notificationService.create(u.getUserId(), NotificationType.REMINDER, NotificationCategory.REMINDER,
+                            "Reminder ⏰", "Time for \"" + u.getTitle() + "\".", u.getId());
+                }
             }
-        }
+
+            // due != fired means a reminder outlived the upgrade it hangs off - a real inconsistency,
+            // and invisible before this line because the loop skips it in silence.
+            log.info("{} {} {}", keyValue("job", REMINDERS_JOB), keyValue("due", due.size()),
```
NaimElijah, line 141: **The comment calls `due != fired` "a real inconsistency", and then logs it at INFO as two unrelated numbers.**

ADR-010's level table, added in this same PR, defines WARN as "degraded, still serving" and INFO as "a state transition". A reminder pointing at an upgrade that no longer exists is the first, not the second — and as written nobody finds it, because finding it means noticing that two fields on an INFO line that is emitted on every non-empty run disagree.

The loop already knows: `due.size() - fired` is the orphan count. One `if (fired < due.size()) log.warn(...)` beside the INFO line turns a fact that is technically present into one somebody will actually see, and matches how the same PR treats the failed push in `StompNotificationPushAdapter`.

backend/src/test/java/com/healthupgrades/common/adapter/out/audit/LoggingAuditTrailTest.java
```diff
@@ -0,0 +1,140 @@
...
+        UUID actor = UUID.randomUUID();
+
+        trail.record(AuditEvent.allowed(AuditAction.REFLECTION_CREATE, actor, UUID.randomUUID()));
+
+        // A reflection's body is the most personal thing this application stores. It cannot reach the
+        // line because AuditEvent has nowhere to hold it (AuditEventTest), and this checks the other
+        // end: what is rendered is only what was passed in.
+        assertThat(single().getFormattedMessage())
+                .contains("action=reflection.create", "actorId=" + actor)
+                .doesNotContain("@", "null");
```
NaimElijah: **This assertion cannot fail.**

The rendered line is `action=reflection.create outcome=ALLOWED resource=REFLECTION actorId=<uuid> resourceId=<uuid>`. A UUID's alphabet is `[0-9a-f-]`, so it can never contain `@`, and `n`/`u`/`l` are not hex digits so it can never contain `null` either. Neither `doesNotContain` argument is reachable by any input this test can produce — it is decoration around the two `contains` checks, not a guard on NFR-6.

The stated intent ("what is rendered is only what was passed in") would bite if it asserted the *shape* of the whole message — that the formatted line is exactly the five expected `key=value` pairs and nothing else — so a sixth field added later fails here.

backend/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Two output formats for one set of log statements: Boot's readable console pattern for a developer,
+  and JSON for a container whose log is read by a machine before a person. The `json-logs` profile
+  selects between them; docker-compose sets it, a local `mvn spring-boot:run` does not.
+
+  See docs/ADRs/ADR-010-structured-logging-a-level-policy-and-audit-as-a-log-stream.md.
```
NaimElijah: **Broken ADR link.** This points at `docs/ADRs/ADR-010-structured-logging-a-level-policy-and-audit-as-a-log-stream.md`, which does not exist — the file added in this PR is `ADR-010-structured-logging-and-a-level-policy.md` (the audit half went to ADR-011).

It matters more here than in an ordinary comment: the block above it is the "LOAD-BEARING" warning about not hand-writing a `<pattern>`, and the ADR is where the reasoning lives. Anyone who follows the link to check before editing this file lands on a 404 and edits anyway.

docker-compose.yml
```diff
@@ -46,8 +58,17 @@ services:
     container_name: healthupgrades-frontend
     ports:
       - "3000:80"
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://localhost/ || exit 1"]
```
NaimElijah: **Two health-checks for one container, and they already disagree.**

`frontend/Dockerfile` now declares `HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 CMD wget -qO- http://localhost/`, and this block re-states the identical probe. Compose's version silently wins, so the image's declaration is decorative here and only applies to whoever runs the image directly.

The backend pair is the same shape but has already drifted: the Dockerfile says `--timeout=5s --retries=3`, compose says `timeout: 10s, retries: 5`. Two sources of truth for the same probe, one of which is unreachable in the only deployment that exists.

ADR-012's stated reason for the image-level `HEALTHCHECK` is "so they are self-describing outside compose" — that is served by the Dockerfile alone. Dropping these two compose `healthcheck:` blocks keeps `depends_on: condition: service_healthy` working (compose reads the image's health status) and leaves one definition per container.

backend/pom.xml
```diff
@@ -111,6 +117,28 @@
...
         </dependency>
 
+        <!--
+          The JSON encoder logback-spring.xml selects under the json-logs profile. A container's log is
+          read by a machine before a person, and the trace id, the level and the audit fields are only
+          queryable if they are fields rather than substrings of a message. See ADR-010.
+        -->
+        <dependency>
+            <groupId>net.logstash.logback</groupId>
+            <artifactId>logstash-logback-encoder</artifactId>
```
NaimElijah, line 127: **Four production classes import this encoder, which contradicts the exit plan ADR-010 commits to.**

`net.logstash.logback.argument.StructuredArguments.keyValue` is imported by `LoggingAuditTrail`, `NotificationScheduler`, `UpgradeOverdueScheduler` and `StompNotificationPushAdapter`. ADR-010 says of the Boot 3.4 upgrade: *"at which point this dependency and most of `logback-spring.xml` are deleted, not migrated."* It cannot be deleted — deleting it breaks four adapters, and Boot's built-in structured logging has no `keyValue` equivalent to swap in.

This is also the only place a *logging backend* reaches into business adapters. Everything else in the codebase talks to SLF4J. A thin wrapper in `common.observability` (`Fields.of("job", name)` returning `Object`) would keep the call sites on one internal API and leave the encoder swap a one-file change — which is what the ADR assumes is already true.

Either add the wrapper, or amend ADR-010 so the upgrade cost is recorded honestly rather than understated.

backend/src/main/java/com/healthupgrades/notification/adapter/in/scheduling/NotificationScheduler.java
```diff
@@ -58,20 +69,29 @@ public class NotificationScheduler {
...
-            boolean loggedToday = !progressQuery.findByUserIdAndDate(userId, today).isEmpty();
-            if (!alreadyNudged && !loggedToday) {
-                notificationService.create(userId, NotificationType.CHECKIN_REMINDER, NotificationCategory.REMINDER,
-                        "Daily check-in ⏳",
-                        "You have " + upgrades.size() + " active upgrade" + (upgrades.size() == 1 ? "" : "s")
-                                + " to track today.", null);
-            }
+            // A count, not a list of user ids: the point is whether the nudge went out at a plausible
+            // volume, and who was nudged is in their own notification list.
+            AtomicInteger nudged = new AtomicInteger();
```
NaimElijah: **`AtomicInteger` for a count that is never concurrent.**

`activeByUser.forEach` on a `HashMap` runs on the calling thread, single-threaded. The atomic is here only to work around "effectively final" in the lambda, and it advertises a thread-safety requirement that does not exist — the next reader has to check whether something parallel is going on.

A plain loop says the same thing without the misdirection:

```java
int nudged = 0;
for (Map.Entry<UUID, List<HealthUpgrade>> entry : activeByUser.entrySet()) {
    ...
    if (!alreadyNudged && !loggedToday) { nudged++; notificationService.create(...); }
}
```

`UpgradeOverdueScheduler` and `dispatchReminders` in this same PR both already count with a plain `int` in a `for` loop — this is the odd one out.

frontend/CLAUDE.md
```diff
@@ -20,6 +20,17 @@ conventions to follow inside `frontend/`.
   `VITE_API_URL` only for a different-origin API). A request interceptor attaches the JWT from
   `localStorage['jwt_token']`; a response interceptor clears the token and redirects to `/login` on 401.
   All `src/api/*.ts` modules call through this client — add new endpoints there, not with raw axios.
+- **A failure is decoded once, in `src/api/apiError.ts`.** `toApiError(thrown)` turns anything a call
```
NaimElijah, line 23: **The rule this bullet states is broken by the one page it matters most on: `LoginPage`.**

`frontend/src/pages/LoginPage.tsx:37` is still:

```tsx
} catch {
  setError('Invalid email or password.');
}
```

A bare `catch {}` that discards the error — exactly what this bullet forbids ("Do not write a hard-coded ... and drop the error") and what `RegisterPage` was fixed for in this PR. Login is also the highest-traffic failure path in the app, and it is unauthenticated, so a backend fault there is the one a user is most likely to hit and least able to describe.

Concrete: `AuthService.login` classifies an `AuthenticationServiceException` (database outage) as `FAILED` and `GlobalExceptionHandler` returns a 500 with a `traceId` on the body — and the user is told their password is wrong, with nothing to quote. The backend work in this PR to make that distinguishable stops at the browser.

`RegisterPage`'s new three lines dropped in here would close it.

NaimElijah:
All fourteen findings addressed in 8e275b2. Every one was legitimate; nothing was argued away.

**The two that reported the opposite of the truth**

- **`ALLOWED` written inside the open transaction.** Confirmed exactly as described — `UpgradeRepositoryAdapter.save` is a bare `jpa.save`, so the version check fires at commit, after the body returned. `LoggingAuditTrail` now defers the `ALLOWED` entry through the same `afterCommit` route `NotificationService` already uses, and records a rollback as `REFUSED`. Refusals stay immediate: they already describe an attempt that did not land, they are the security-relevant half, and a process that dies before commit should not take them with it. `AuditCommitIT` pins it against a real transaction manager — and I verified it by disabling the deferral, which fails it.
- **`Error` counted as `ok`.** Took the suggestion. Worth noting: my first two attempts to apply this fix aborted on a mismatched javadoc anchor before writing, so the code stayed unpatched — and the new test is what caught that.

**Knock-on you flagged:** `AuditOutcome.of`'s `OptimisticLockException` branch stays, but the test is renamed off `GivenALostConcurrentEdit_…` and both now say plainly that a commit-time clash never reaches the classifier and is handled on the adapter's rollback path. Keeping the function total over the domain's four "no" exceptions is deliberate; implying a live scenario was not.

**One I want to be explicit about not doing:** the parenthetical about `Error` in `AuditTrail.recording`. An `Error` there leaves no entry — an absence, not a lie — and catching `Error` to write a log line during an `OutOfMemoryError` is as likely to throw as to help. `JobMetrics` was different because it actively recorded the wrong answer. Left as is.

**The rest**

`management.metrics.tags` added (Boot's own metadata settles it — `key-values` is "every observation", `tags` is "every meter"); `record()` guarded so a failing meter cannot replace a business exception; `AuthService` records after the token exists; `recordingCreation` so a create names what it created; the orphaned reminder is now `WARN`; `LoginPage` keeps the vague message for 401 only; the ADR link, the duplicated healthcheck, the `AtomicInteger` and ADR-010's "deleted, not migrated" claim all corrected.

**The three weak tests** are the findings I was most glad of — each certified a property it could not fail on. Fixed as exact-line, per-family, and real `AxiosHeaders` respectively. `LoginPage`'s stubs rejected with `new Error('401')`, which carries no status and so could not model the distinction the page now draws; they now reject with a real axios response, and there is a new test that a 500 must not be reported as a wrong password.

Verified: `mvn clean verify` 465 unit + 39 integration; vitest 126; tsc, eslint `--max-warnings 0`, `check:colours` and the diagram drift check all clean.

PR #69, opened 2026-08-26
docs: give the README a deliberate visual pass

NaimElijah:
Follows [#67](https://github.com/NaimElijah/UpHealther/pull/67), which fixed the README's structure.
This one is about how the page *looks*. No content claims changed; every command, version and number
is the same one #67 verified.

## Problem

The content was right and the page was flat. No centred header, no callouts, no horizontal rhythm,
and badges carrying shields.io's default style — so the first screen read as a wall of grey text. A
reader decides whether to keep reading in roughly five seconds, and a page that looks neglected
suggests code that is.

## What was done

- **Centred the header block** — name, description, badges, doc links — inside `<div align="center">`
  with the blank lines that keep the markdown rendering. It is the only centred content in the file;
  body text stays left-aligned, because centred prose loses the line-start the eye reads down.
- **Made the badges a set:** one shared `style=flat-square`, ordered by volatility. Dropped a
  hand-written `tests 630 passing` and a static `Java 21` badge I had drafted — neither carries live
  state, which is the bar. Three remain: CI, last commit, licence. Added the API reference to the
  link row.
- **Three GitHub alerts**, well separated, at the three points a reader must not misread: the
  medical disclaimer, "nothing is deployed", and source-available-not-open-source.
- **Two horizontal rules** — after the header, before the licence — so the page has movements
  instead of one undifferentiated column.
- **Right-aligned the version column** so the numbers compare down the column, and cut two table
  cells that had become paragraphs.
- **Every code line now fits in 88 characters**, so no block scrolls horizontally. Eight did; the
  project tree accounted for six of them.
- **No emoji, deliberately.** The policy is all headings or none, and none suits a document whose
  audience is reading it to judge the engineering.

## Trade-offs accepted

- **409 lines, against the 200-400 target.** The design pass added roughly twenty lines, paid for by
  folding the login round-trip into a sentence and the cron table into one. This project has an
  unusually large legitimate surface — nine bounded contexts, twelve ADRs, a full observability
  story — and the depth is already linked out to `docs/`. Trimming the last nine lines would cost
  content from Features or Testing, the two most-read sections.
- **Five kinds of visual device in play** (badges, alerts, tables, a diagram, one `<details>`),
  where my own guidance says four. The `<details>` is the only collapsed content and it holds real
  troubleshooting, so it stayed.
- **`<div align>` and the alerts are GitHub-specific.** This repository is only published on GitHub,
  so the trade is free here; it would not be for a package that also renders on npm.

## A correction inside the change

The README's shortened create-upgrade example previously said the response "is reproduced in full in
docs/api.md", but the two requests differed once I shortened the README's payload to fit the code
fence. It now says a complete request-and-response *pair* is there, which is what is true.

## How it was verified

`python docs/architecture/arch-diagrams/generate.py --check` — diagrams up to date. The rest is
structural, checked mechanically against the source:

| Check | Result |
|---|---|
| Relative links and heading anchors | all resolve, both files |
| Badge style consistency | one (`flat-square`) |
| Link separator | one, 27 uses |
| Heading case | sentence case throughout |
| Centred blocks | 1 — the header only |
| Mermaid hardcoded `style` lines | 0, so the diagram reads in both themes |
| Code lines over 88 chars | 0 |
| Table columns | 4 maximum |
| Fences, `<details>`, `<div>` | all balanced |

**Not verified:** the rendered page has not been viewed on GitHub — the checks above are on the
source. Worth a glance at the Files tab before merge, particularly the centred header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,21 @@
+<div align="center">
+
 # UpHealther
 
-Plan a health change, then find out whether it actually happened — a tracker for habits,
+**Plan a health change, then find out whether it actually happened** — a tracker for habits,
 experiments, goals and one-off health actions, for people who want structure without a coach.
 
-[![CI](https://github.com/NaimElijah/UpHealther/actions/workflows/ci.yml/badge.svg)](https://github.com/NaimElijah/UpHealther/actions/workflows/ci.yml)
-[![Licence: source-available](https://img.shields.io/badge/licence-source--available-blue)](LICENSE)
+[![CI](https://img.shields.io/github/actions/workflow/status/NaimElijah/UpHealther/ci.yml?branch=main&style=flat-square&label=CI)](https://github.com/NaimElijah/UpHealther/actions/workflows/ci.yml)
+[![Last commit](https://img.shields.io/github/last-commit/NaimElijah/UpHealther?style=flat-square)](https://github.com/NaimElijah/UpHealther/commits/main)
+[![Licence](https://img.shields.io/badge/licence-source--available-blue?style=flat-square)](LICENSE)
 
 [Requirements](docs/requirements/requirements.md) ·
 [Architecture](docs/architecture/architecture.md) ·
 [Diagrams](docs/architecture/arch-diagrams/README.md) ·
+[API](docs/api.md) ·
 [Decisions](docs/ADRs/)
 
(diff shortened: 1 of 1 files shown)


NaimElijah:
Reopening to trigger CI — the workflow does not fire on the initial PR from a freshly pushed branch.

Issue #59, opened 2026-08-25
ci: the backend has no dependency vulnerability audit, and the tool that blocked it is no longer the only option


NaimElijah:
The frontend audits what it ships on every push:

```yaml
- name: Audit shipped frontend dependencies for known vulnerabilities
  run: cd frontend && npm audit --omit=dev --audit-level=high
```

The backend audits nothing. `ci.yml` says so in a comment, and ADR-002 records why:

> **The backend has no vulnerability audit yet.** OWASP dependency-check was added, run, and removed
> again: it cannot populate its database without an `NVD_API_KEY`, and fails outright without one. A step
> that always fails, or one marked `continue-on-error`, is the same decorative guard this record exists to
> remove. Enabling it means setting the repository secret first.

That reasoning is sound and it is about **one tool**. The conclusion drawn from it — no backend audit —
does not follow any more, because there are scanners that need no key at all.

### What has changed since ADR-002

| Option | Key needed | Notes |
|---|---|---|
| **OSV-Scanner** | none | Reads `pom.xml` / the resolved tree against the OSV database. Runs as an action or a binary; fails the build on a finding. |
| **Trivy** (`fs` scan) | none | Same shape, broader remit — it also scans the images, which is the other half of item 10 of #51. |
| **GitHub Dependabot alerts** | none | Repository setting, not a CI step. Free on a public repository, which this is. Alerts and updates, but nothing that reds a PR. |
| **`actions/dependency-review-action`** | none | PR-scoped: fails when a PR *introduces* a vulnerable dependency. Complements a full scan rather than replacing it. |
| **OWASP dependency-check** | `NVD_API_KEY` | The one ADR-002 tried. Still the option with the secret. |

### Proposed work

1. Pick one scanner for the backend and one severity threshold that fails the build. `--audit-level=high`
   is the precedent the frontend already sets; matching it keeps one rule across both modules.
2. Add the step to the `backend` job. Not `continue-on-error` — ADR-002 is explicit that a guard which
   cannot fail is worse than none.
3. Turn on Dependabot alerts, and decide separately whether to accept its PRs — that is a maintenance
   cadence question, not a security one.
4. **A new ADR** recording the choice, superseding ADR-002's known-gap paragraph, with the reason the
   rejected options were rejected. ADR-002 is accepted and append-only, so this is a new record that
   points at it, not an edit.
5. Close the `requirements.md` section 6 entry, and give NFR-12 the backend half of its sentence.

### Worth noting

[...]

PR #72, opened 2026-08-28
fix: enforce at the boundary what was only being assumed

### LINKED ISSUE #53: fix: an over-long title or name returns 500 instead of 400
Every request DTO bounds its numbers and none of its strings. `UpgradeRequest.title` is `@NotBlank` and
nothing more (`backend/.../upgrade/adapter/in/web/UpgradeRequest.java:30`); so is `HealthAreaRequest.name`
(`HealthAreaRequest.java:19`). The columns behind them are not unbounded:

| Column | Type | Bound at the boundary? |
|---|---|---|
| `health_upgrades.title` | `VARCHAR(255)` | no |
| `health_areas.name` | `VARCHAR(255)` | no |
| `health_areas.icon` | `VARCHAR(100)` | no |
| `health_areas.color` | `VARCHAR(50)` | no |
| `progress_entries.unit` | `VARCHAR(100)` | no |
| `tracking_configs.target_unit` | `VARCHAR(100)` | no |
| `users.name`, `users.email` | `VARCHAR(255)` | no |

A 256-character title therefore passes validation, passes the service, and fails at the flush as a
`DataIntegrityViolationException`. `GlobalExceptionHandler` maps no such type — the only handler that
catches it is `@ExceptionHandler(Exception.class)` (`GlobalExceptionHandler.java:231`), which answers 500
and logs a stack trace.

**Which requirement this contradicts.** NFR-7: *"400 invalid input (a failed constraint, an unbindable
body, a parameter that will not convert)"*. An over-long field is invalid input — the client can fix it,
and a 500 tells them nothing they can act on. It is the same shape as #19 and #22, both closed.

**Traced through the code, not executed.** The first step is the failing test that proves it, per the
repo's rule that a bug fix starts with a reproduction.

### Proposed work

1. A controller test posting an over-long title and asserting the status. That test *is* the bug report.
2. `@Size(max = ...)` on every string field of every request DTO, each bound taken from the column it
   lands in, so the refusal happens at the boundary and names the field that failed.
3. Decide whether `DataIntegrityViolationException` should also be mapped as a backstop — a 400 or a 422
   rather than a 500 — for the constraints a DTO annotation cannot express. `DuplicateProgressException`
   already shadows one such constraint (BR-6); this would be the general case behind it.
4. Update `requirements.md` if the outcome adds a rule worth naming.

### Worth deciding in the same pass

`description`, `motivation`, `success_criteria`, `note`, `what_worked`, `what_did_not_work` and
`next_adjustment` are `TEXT` — unbounded in the database, so no 500, but also no ceiling on what a single
request can store. Whether those get a bound is a product call, not a defect.

### LINKED ISSUE #54: fix: the API accepts a one-character password; only the browser asks for six
`RegisterRequest.password` is `@NotBlank` and nothing more (`AuthController.java:81`). `RegisterPage`
refuses anything shorter than six characters (`RegisterPage.tsx:37`), and its own comment says the rules
there *"are for immediate feedback"* — which reads as though a server-side rule sits behind it. None does.
`POST /api/auth/register` with `"password": "a"` is accepted and the account is created.

**Which principle this contradicts.** Validation belongs at the boundary; a rule enforced only in the
client is not enforced. Any caller that is not the frontend bypasses it entirely, and the frontend is not
a trusted component — it runs on the caller's machine.

It also widens #56: the online search space starts
at one character.

### Proposed work

1. `AuthServiceTest` / `AuthControllerTest`: a password below the minimum is refused 400; one at the
   minimum is accepted.
2. `@Size(min = ..., max = 72)` on `RegisterRequest.password`. The upper bound is not cosmetic — BCrypt
   truncates at 72 bytes, so a 100-character password and its first 72 characters are the same password.
   Refusing the input is honest; silently truncating it is not.
3. Keep the frontend's minimum in step with the server's, and leave it in place — immediate feedback is
   still worth having.
4. A line in `requirements.md` §2.1, since this becomes a rule the system enforces.

### Open, and part of the same decision

Whether the minimum stays at six. Six is what the UI has always asked for and is below every current
recommendation. Raising it is free for new accounts and cannot be applied to existing ones without a
password-change endpoint — which does not exist; see #57.

### LINKED ISSUE #60: a11y: a dialog does not trap focus, so aria-modal cannot honestly be claimed
`requirements.md` section 6 records this. Filed so it is tracked as work rather than only as a caveat.

`Modal.tsx` says it plainly in its own doc comment:

> Deliberately no `aria-modal`. It would tell assistive technology that everything outside the dialog is
> inert, and nothing here makes that true: no focus trap, so Tab walks straight out into the page behind,
> and focus is neither moved in on open nor restored on close. Claiming it would leave a [...]

That judgement is right — an honest `role="dialog"` beats a false `aria-modal`. What is missing is the
work that would make the claim true.

**What the dialog does have today:** `role="dialog"`, `aria-labelledby` pointing at its title (FR-40,
`Modal.test.tsx`), Escape to close, a backdrop that closes, and a body that scrolls while the heading
stays put (FR-39).

**What it does not have:** focus moved into the dialog on open, focus restored to the trigger on close,
Tab and Shift+Tab confined to the dialog, and the inertness of the page behind it.

### Two ways to close it

**A — native `<dialog>` with `showModal()`.** The browser provides the focus trap, the inertness, the
backdrop and the Escape handling. Least code, and the behaviour is the platform's rather than ours.
ADR-005 records why it was not done at the time. Costs: the styling hooks change (`::backdrop`, the
default UA styles), and jsdom's `<dialog>` support is partial — some of the current
`Modal.test.tsx` assertions will need rewriting against a different DOM shape.

**B — keep the div and implement the trap by hand.** A ref to the first focusable element, a keydown
handler that cycles Tab at the boundaries, a stored trigger to restore to, and `inert` or
`aria-hidden` on the app root. More code, fully testable in jsdom, and it is a well-worn pattern that is
still easy to get subtly wrong.

A is the better answer if the test cost is acceptable; B is the one that keeps the existing tests.

### Done means

`aria-modal="true"` is set and true, tests cover focus-on-open, focus-restored-on-close and the Tab cycle
at both boundaries, the section 6 entry is gone, and FR-40 grows the part of the contract it is now
allowed to claim.

NaimElijah:
## The problem

Three defects that share a shape: a rule the system appeared to have, and did not.

- **#53** — every request DTO bounded its numbers and none of its strings. A 256-character title
  passed validation, passed the service and failed at the flush as a `DataIntegrityViolationException`.
  Nothing maps that type, so it reached the catch-all and came back **500** for input the caller could
  have fixed. That contradicts NFR-7.
- **#54** — `RegisterRequest.password` was `@NotBlank` and nothing more, so `POST /api/auth/register`
  with `"password": "a"` created the account. The register page has always asked for a minimum and its
  comment described the rule as immediate feedback, which reads as though a server rule sat behind it.
  None did, and the frontend runs on the caller's machine.
- **#60** — `Modal` declined `aria-modal` on purpose because nothing made it true: no focus trap, and
  focus neither moved in on open nor restored on close.

## The approach

**Bounds come from the columns, not from taste.** Each `@Size(max)` mirrors the `VARCHAR(n)` the field
lands in, read out of `V1__init_schema.sql`. The two count differently — `@Size` counts UTF-16 code
units where Postgres counts characters — which makes the boundary *stricter* for an astral-plane
character such as an emoji, and never looser. That is the safe direction and it is stated where the
constants are declared.

**The password's upper bound is BCrypt's, not policy.** `BCryptPasswordEncoder` in Spring Security
6.2.4 guards only against null — checked in the resolved jar's bytecode, not assumed — so anything past
72 bytes is silently dropped and a longer password becomes indistinguishable from its own prefix.
Refusing it is honest; truncating it quietly is not.

**`aria-modal` needed more than a Tab handler.** The overlay is portalled to `<body>` so "the page
behind" is one list of nodes rather than an ancestor's siblings at six depths, and those nodes are
marked `inert`. `inert` rather than `aria-hidden` is the deciding choice, and it is not about the
browser: Testing Library treats an `aria-hidden` ancestor as non-existent for role queries and knows
nothing about `inert`, so `aria-hidden` would leave the next page test that opens a dialog unable to
query the page underneath it. Recorded in **ADR-013**, which supersedes the part of ADR-005 that had
this as an open question.

## The trade-offs accepted

- **The password minimum now lives in two places** — `AuthController.PASSWORD_MIN` and `RegisterPage` —
  with nothing keeping them in step. A contract test in the shape of `FrontendEnumContractTest` would
  fix that and is more machinery than one integer earns. Both say so.
- **`TEXT`-backed fields are left unbounded.** Nothing downstream rejects them, so there is no 500 to
  prevent; how long a reflection may be is a product decision. Section 6 records it, along with the
  wider question of whether `DataIntegrityViolationException` should be mapped as a backstop at all.
- **Login is left unbounded** deliberately: a wrong-length password there is a failed credential and
  already answers 401, and bounding it would leak the rule to an unauthenticated caller.
- **The dialog's inertness is not verifiable here.** jsdom implements `inert` not at all, so the tests
  pin that the attribute is set and cleared on the right nodes and stop there. Section 6 keeps a
  narrower entry for that, plus two smaller residuals: focus falls back to `<body>` when the trigger
  unmounts with the dialog, and a toast raised over an open dialog goes inert with the page.
- **The demo password changed** from `demo123` to `demo1234`, because seven characters would no longer
  meet the rule the API now enforces. `V6` does it as an `UPDATE`; `V2` and `V3` are untouched, since
  Flyway checksums a migration file whole and correcting even their comments would break validation on
  an already-migrated database.

## What was done

| Commit | |
|---|---|
| `2cf55ac` | `@Size(max)` on every string field with a bounded column, plus six boundary tests |
| `58511cc` | `@Size(min = 8, max = 72)` on the registration password, three tests, browser rule raised to match |
| `6529f97` | `V6` raises the seeded demo password to the new minimum |
| `efb5143` | Portal, refcounted `inert`, Tab trap, `aria-modal`, twelve tests, ADR-013 |

Requirements gain **BR-16** and **BR-17**; **FR-40** grows to the contract the dialog can now honestly
claim. The traceability counts in the header were recounted, not adjusted by arithmetic.

## How it was verified

- `mvn verify` — 475 unit/slice/architecture tests and 39 integration tests green, including `V6`
  applying cleanly through Flyway against a Testcontainers Postgres.
- `npm run lint`, `npm run check:colours`, `npm run test` (140), `npm run build` — all clean.
- **Each new test was seen to fail first.** The six boundary tests reproduced the 500 before the bounds
  went in; both password tests failed before the annotation did.
- **The three subtle branches in the focus trap were mutation-checked**: removing the refcount, the
  overlay skip, or `preventDefault` fails exactly the tests written for each and nothing else.
- **End-to-end against `docker-compose`**: `demo1234` logs in (200) and `demo123` no longer does (401);
  a 256-character title answers `400 {"title":"size must be between 0 and 255"}`; a one-character
  password answers `400 {"password":"size must be between 8 and 72"}`.

Closes #53, closes #54, closes #60.

### DIFF
--- a/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
+++ b/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
@@ -25,17 +25,17 @@ import org.springframework.web.bind.annotation.*;
 public class AuthController {
 
     /** Mirrors {@code users.name VARCHAR(255)}; see BR-16 for why the bound is here. */
-    static final int NAME_MAX = 255;
+    public static final int NAME_MAX = 255;
 
     /** Mirrors {@code users.email VARCHAR(255)}. */
-    static final int EMAIL_MAX = 255;
+    public static final int EMAIL_MAX = 255;
 
     /**
      * The shortest password the API accepts. Unlike the bounds above this mirrors no column - only the
      * hash is stored, and a BCrypt hash is always 60 characters - so it is policy, and it is enforced
      * here because a rule the browser alone applies is not enforced at all. BR-17.
      */
-    static final int PASSWORD_MIN = 8;
+    public static final int PASSWORD_MIN = 8;
 
     /**
      * Where BCrypt stops reading. {@code BCryptPasswordEncoder} guards only against null, so anything
@@ -44,7 +44,7 @@ public class AuthController {
      * {@code @Size} counts UTF-16 code units and BCrypt counts bytes, so a password of multi-byte
      * characters can pass this bound and still be truncated.
      */
-    static final int PASSWORD_MAX = 72;
+    public static final int PASSWORD_MAX = 72;
 
     private final AuthService authService; // application service
 
--- a/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
+++ b/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
@@ -30,6 +30,22 @@ public class AuthController {
     /** Mirrors {@code users.email VARCHAR(255)}. */
     static final int EMAIL_MAX = 255;
 
+    /**
+     * The shortest password the API accepts. Unlike the bounds above this mirrors no column - only the
+     * hash is stored, and a BCrypt hash is always 60 characters - so it is policy, and it is enforced
+     * here because a rule the browser alone applies is not enforced at all. BR-17.
+     */
+    static final int PASSWORD_MIN = 8;
+
+    /**
+     * Where BCrypt stops reading. {@code BCryptPasswordEncoder} guards only against null, so anything
+     * past this is silently dropped and a longer password becomes indistinguishable from its own
+     * prefix; refusing it is honest where truncating it is not. Note the mismatch this cannot close:
+     * {@code @Size} counts UTF-16 code units and BCrypt counts bytes, so a password of multi-byte
+     * characters can pass this bound and still be truncated.
+     */
+    static final int PASSWORD_MAX = 72;
+
     private final AuthService authService; // application service
 
     /**
@@ -85,7 +101,7 @@ public class AuthController {
     public record RegisterRequest(
             @NotBlank @Size(max = NAME_MAX) String name,
             @NotBlank @Email @Size(max = EMAIL_MAX) String email,
-            @NotBlank String password
+            @NotBlank @Size(min = PASSWORD_MIN, max = PASSWORD_MAX) String password
     ) {}
 
     /** Login request body. */
--- a/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
+++ b/backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
@@ -8,6 +8,7 @@ import com.healthupgrades.user.domain.model.User;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.Email;
 import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -22,6 +23,12 @@ import org.springframework.web.bind.annotation.*;
 @RequestMapping("/api/auth")
 @RequiredArgsConstructor
 public class AuthController {
+
+    /** Mirrors {@code users.name VARCHAR(255)}; see BR-16 for why the bound is here. */
+    static final int NAME_MAX = 255;
+
+    /** Mirrors {@code users.email VARCHAR(255)}. */
+    static final int EMAIL_MAX = 255;
 
     private final AuthService authService; // application service
 
@@ -76,8 +83,8 @@ public class AuthController {
 
     /** Registration request body. */
     public record RegisterRequest(
-            @NotBlank String name,
-            @NotBlank @Email String email,
+            @NotBlank @Size(max = NAME_MAX) String name,
+            @NotBlank @Email @Size(max = EMAIL_MAX) String email,
             @NotBlank String password
     ) {}
 
--- a/backend/src/main/java/com/healthupgrades/healtharea/adapter/in/web/HealthAreaRequest.java
+++ b/backend/src/main/java/com/healthupgrades/healtharea/adapter/in/web/HealthAreaRequest.java
@@ -1,6 +1,7 @@
 package com.healthupgrades.healtharea.adapter.in.web;
 
 import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.Size;
 
 /**
  * Request body for creating or replacing a health area.
@@ -8,7 +9,7 @@ import jakarta.validation.constraints.NotBlank;
  * <p>Used by both {@code POST} and {@code PUT}: an update is a full replacement, so every optional
  * field left out is stored as null rather than kept at its previous value.
  *
- * @param name        display name; required and non-blank
+ * @param name        display name; required, non-blank and no longer than its column
  * @param description free-text note, optional
  * @param priority    ordering hint, optional
  * @param icon        the glyph to draw for the area, usually an emoji; optional. Not an icon-font key
@@ -16,9 +17,18 @@ import jakarta.validation.constraints.NotBlank;
  * @param color       colour token, optional
  */
 public record HealthAreaRequest(
-        @NotBlank String name,
+        @NotBlank @Size(max = NAME_MAX) String name,
         String description,
         Integer priority,
-        String icon,
-        String color
-) {}
+        @Size(max = ICON_MAX) String icon,
+        @Size(max = COLOR_MAX) String color
+) {
+    /** Mirrors {@code health_areas.name VARCHAR(255)}; see BR-16 for why the bound is here. */
+    static final int NAME_MAX = 255;
+
+    /** Mirrors {@code health_areas.icon VARCHAR(100)}. */
+    static final int ICON_MAX = 100;
+
+    /** Mirrors {@code health_areas.color VARCHAR(50)}, the narrowest of the three. */
+    static final int COLOR_MAX = 50;
+}
--- a/backend/src/main/java/com/healthupgrades/healtharea/adapter/in/web/HealthAreaRequest.java
+++ b/backend/src/main/java/com/healthupgrades/healtharea/adapter/in/web/HealthAreaRequest.java
@@ -24,11 +24,11 @@ public record HealthAreaRequest(
         @Size(max = COLOR_MAX) String color
 ) {
     /** Mirrors {@code health_areas.name VARCHAR(255)}; see BR-16 for why the bound is here. */
-    static final int NAME_MAX = 255;
+    public static final int NAME_MAX = 255;
 
     /** Mirrors {@code health_areas.icon VARCHAR(100)}. */
-    static final int ICON_MAX = 100;
+    public static final int ICON_MAX = 100;
 
     /** Mirrors {@code health_areas.color VARCHAR(50)}, the narrowest of the three. */
-    static final int COLOR_MAX = 50;
+    public static final int COLOR_MAX = 50;
 }
--- a/backend/src/main/java/com/healthupgrades/upgrade/adapter/in/web/UpgradeRequest.java
+++ b/backend/src/main/java/com/healthupgrades/upgrade/adapter/in/web/UpgradeRequest.java
@@ -4,6 +4,7 @@ import com.healthupgrades.upgrade.domain.model.Difficulty;
 import com.healthupgrades.upgrade.domain.model.UpgradeType;
 import jakarta.validation.constraints.NotBlank;
 import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Size;
 
 import java.time.LocalDate;
 import java.util.UUID;
@@ -16,7 +17,7 @@ import java.util.UUID;
  * belongs to {@code /plan} and {@code /reschedule}.
  *
  * @param areaId           health area to file it under, optional
- * @param title            short name; required and non-blank
+ * @param title            short name; required, non-blank and no longer than its column
  * @param description      longer explanation, optional
  * @param type             what kind of change this is; required
  * @param difficulty       how demanding it is, optional
@@ -27,7 +28,7 @@ import java.util.UUID;
  */
 public record UpgradeRequest(
         UUID areaId,
-        @NotBlank String title,
+        @NotBlank @Size(max = TITLE_MAX) String title,
         String description,
         @NotNull UpgradeType type,
         Difficulty difficulty,
@@ -35,4 +36,7 @@ public record UpgradeRequest(
         LocalDate targetEndDate,
         String motivation,
         String successCriteria
-) {}
+) {
+    /** Mirrors {@code health_upgrades.title VARCHAR(255)}; see BR-16 for why the bound is here. */
+    static final int TITLE_MAX = 255;
+}
--- a/backend/src/main/java/com/healthupgrades/upgrade/adapter/in/web/UpgradeRequest.java
+++ b/backend/src/main/java/com/healthupgrades/upgrade/adapter/in/web/UpgradeRequest.java
@@ -37,6 +37,11 @@ public record UpgradeRequest(
         String motivation,
         String successCriteria
 ) {
-    /** Mirrors {@code health_upgrades.title VARCHAR(255)}; see BR-16 for why the bound is here. */
-    static final int TITLE_MAX = 255;
+    /**
+     * Mirrors {@code health_upgrades.title VARCHAR(255)}; see BR-16 for why the bound is here.
+     *
+     * <p>Public because {@code ColumnBoundContractTest} reads it from another package to check it
+     * still agrees with the migration, which is the only thing keeping the two numbers in step.
+     */
+    public static final int TITLE_MAX = 255;
 }
--- a/docs/requirements/requirements.md
+++ b/docs/requirements/requirements.md
@@ -2,7 +2,7 @@
 
 What UpHealther must do. This document records the requirements the project **currently meets** —
 each one is implemented, and the **test** that enforces it is named, so a claim here can be checked
-rather than trusted. Eighty of the eighty-seven entries below name a test — sixty-one distinct
+rather than trusted. Eighty-one of the eighty-eight entries below name a test — sixty-one distinct
 test classes and files between them. Four of the remaining seven name the command, workflow or script
 that *is* the check (NFR-11, NFR-12, NFR-13, NFR-18). The last three — FR-39, NFR-19 and NFR-20 — are
 verified by hand and say so, because each is about a rendered width, a colour or an overflow, and jsdom
@@ -133,6 +133,14 @@ nothing is shared between accounts.
 | BR-13 | Reflections are append-only — there is no edit or delete path | `ReflectionServiceTest` (asserted against the public surface), `ReflectionControllerTest` |
 | BR-14 | Concurrent edits to an upgrade are refused rather than silently merged | `UpgradePersistenceIT`, `GlobalExceptionHandlerTest` |
 | BR-15 | A record is visible only to its owner; another user's record is reported as absent, never as forbidden | `UpgradePersistenceIT`, `HealthAreaPersistenceIT`, `ProgressEntryPersistenceIT`, and every `*ControllerTest` |
+| BR-16 | A field stored in a bounded column is refused at the boundary when it exceeds that bound, and the response names the field | `UpgradeControllerTest`, `HealthAreaControllerTest`, `ProgressControllerTest`, `TrackingConfigControllerTest` |
+
+BR-16 exists because the alternative is a 500. Each bound is taken from the column the field lands in
+(`V1__init_schema.sql`), so the two cannot drift apart in the direction that matters: `@Size` counts
+UTF-16 code units where `VARCHAR(n)` counts characters, which makes the boundary *stricter* than the
+column for an astral-plane character such as an emoji, and never looser. Fields stored in `TEXT`
+columns are deliberately unbounded — nothing rejects them downstream, so a ceiling there is a product
+decision rather than a defect, and §6 records it as open.
 
 ---
 
@@ -222,6 +230,16 @@ Undecided, and owned by the repository owner.
   records why that was not done here.
 - **The backend has no dependency vulnerability audit.** OWASP dependency-check needs an `NVD_API_KEY`
   secret; ADR-002 records why a check that cannot fail was judged worse than none.
+- **No ceiling on the free-text fields.** `description`, `motivation`, `successCriteria`, `note`,
+  `whatWorked`, `whatDidNotWork` and `nextAdjustment` are `TEXT`, so BR-16 leaves them alone: nothing
+  downstream rejects them and there is no 500 to prevent. What is left is that a single request can
+  store as much prose as the servlet container will accept. How long a reflection may be is a product
+  decision, not a defect, and nobody has taken it.
+- **Whether `DataIntegrityViolationException` should be mapped at all.** BR-16 removes the known route
+  to it, but it stays unmapped, so any constraint a DTO annotation cannot express still surfaces as a
+  500. Mapping it centrally is not one decision but several — a unique violation, a foreign-key
+  violation and a not-null violation do not deserve the same status, and `DuplicateProgressException`
+  already shadows the first of them.
 - **`UpgradeType.PROTOCOL` is deprecated but retained** for rows that may already carry it. Removing it
   needs confirmation that no stored row uses it.
 - **No governing jurisdiction is named in the licence** — ADR-003 flags this as the first thing to add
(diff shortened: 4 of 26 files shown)


backend/src/main/java/com/healthupgrades/upgrade/adapter/in/web/UpgradeRequest.java
```diff
@@ -27,12 +28,15 @@
  */
 public record UpgradeRequest(
         UUID areaId,
-        @NotBlank String title,
+        @NotBlank @Size(max = TITLE_MAX) String title,
```
NaimElijah, line 31: **The 400 this bound produces is thrown away by the UI, so the caller still cannot fix the input.**

`UpgradeBacklogPage.handleCreate` is `await createMutation.mutateAsync(form)` with no `catch` (line 72); `HealthAreasPage` does the same at lines 56 and 62, and `UpgradeDetailsPage` uses bare `.mutate(...)` with no `onError` for progress, reflection and tracking config. A 256-character title now rejects with `400 {"title":"size must be between 0 and 255"}`, the promise rejects unhandled, `formError` is never set, and the modal stays open showing nothing — the same nothing the 500 produced. The inputs carry no `maxLength` either, so the browser does not stop it first.

The PR's premise is "a server fault reported for input the caller could have corrected". The correction never reaches the caller: the fix stops at the API boundary. Rendering `toApiError(thrown).fieldErrors` into the field it names (the `Input`s already take an `error` prop) is what closes it.

frontend/src/pages/RegisterPage.tsx
```diff
@@ -34,7 +42,10 @@ const RegisterPage: React.FC = () => {
     setError('');
     if (!name || !email || !password) { setError('Please fill in all fields.'); return; }
     if (password !== confirm) { setError('Passwords do not match.'); return; }
-    if (password.length < 6) { setError('Password must be at least 6 characters.'); return; }
+    if (password.length < PASSWORD_MIN_LENGTH) {
```
NaimElijah, line 45: **Only the minimum is mirrored here; the maximum is not, and the 400 it produces is unreadable.**

`PASSWORD_MAX = 72` has no counterpart on this page, so a password longer than 72 characters — routine for anyone pasting a password manager's passphrase — passes this check and reaches the API. The API answers `400` with `message: "Validation failed"` and the actual reason only in `fieldErrors.password`. Line 59 renders `failure.message` and drops `fieldErrors`, so the user sees

> Validation failed (reference 7f3a…)

and has no way to know the password was too long. Same for a name or an email over 255. `frontend/CLAUDE.md` says a failure is decoded once into `{ status, message, fieldErrors, traceId }` — this call site decodes it and then discards half of it.

Either mirror `PASSWORD_MAX` here too, or render `fieldErrors` when the response carries them.

frontend/src/components/ui/Modal.tsx
```diff
@@ -49,15 +104,75 @@ const Modal: React.FC<ModalProps> = ({ isOpen, onClose, title, children }) => {
     return () => document.removeEventListener('keydown', handler);
```
NaimElijah, line 128: **Escape still closes every open dialog at once — the one place two dialogs do fight over a key.**

This listener is on `document` and is bound once per open modal, so a single Escape runs every open modal's `onClose`. The doc comment above used to claim "several on one page cannot fight over the key"; this PR removed the claim without removing the fight, while adding a refcount and an `aria-modal` claim that both assume stacking works.

With two dialogs open, Escape closes the top one *and* the one behind it, `openModals` falls to 0, and the page behind is released from `inert` in the same tick. `GivenTwoOpenModals_WhenOneCloses_ThenTheContentBehindStaysInert` only passes because the outer modal's `onClose` is `() => {}`; give it a real handler and the assertion fails.

The Tab trap already solved exactly this by binding to the dialog rather than to `document` — Escape can do the same (the container is focused and holds focus), or only the topmost modal can act.

frontend/src/components/ui/Modal.tsx
```diff
@@ -13,6 +14,47 @@ interface ModalProps {
...
+ * which has to be set in the JSX rather than registered from an effect — React inserts both portals
+ * during the mutation phase and runs neither effect until afterwards, so an effect-time registry
+ * loses that race and the first dialog marks the second one inert.
+ */
+let openModals = 0;
+let inerted: Element[] = [];
+
+const acquireInert = () => {
+  if (++openModals > 1) return;
+  inerted = Array.from(document.body.children).filter(
```
NaimElijah, line 64: **The inert set is snapshotted once, so anything appended to `<body>` after the first dialog opens is never inerted — and `aria-modal="true"` is then a false claim about it.**

`acquireInert` reads `document.body.children` on the 0→1 transition only. A node added to `<body>` afterwards is outside `inerted`: it stays interactive and in the accessibility tree while the dialog claims everything behind it is not. Conversely a node that merely happens to exist *before* the dialog opens is inerted whether that was wanted or not.

Nothing else portals to `document.body` today, so this is latent rather than live — but the invariant ("any element under `<body>` that is not a dialog overlay must carry `data-modal-overlay`, or accept being inerted") lives only inside this file, and the next `createPortal(..., document.body)` gets one of the two behaviours purely by mount timing. `ToastContainer` inside `#root` is already the acknowledged half of this.

A `MutationObserver` on `document.body` while `openModals > 0`, or re-running the walk from whichever child mounts, removes the timing dependence.

frontend/src/components/ui/Modal.tsx
```diff
@@ -49,15 +104,75 @@ const Modal: React.FC<ModalProps> = ({ isOpen, onClose, title, children }) => {
...
+    if (items.length === 0) {
+      // Unreachable through this component's own markup — the close button is always rendered and
+      // never disabled — but focus still has to land somewhere if a future body removes it.
+      e.preventDefault();
+      dialog.focus();
+      return;
+    }
+
+    const index = items.indexOf(document.activeElement as HTMLElement);
+    const wrapToLast = e.shiftKey && index <= 0;
```
NaimElijah: **`index <= 0` conflates "focus is on the first control" with "focus is not in the list at all", and `FOCUSABLE` misses several natively focusable elements.**

`items.indexOf(...)` returns `-1` both when focus is on the dialog container (intended, and documented above) *and* when focus is on any element `FOCUSABLE` does not match. The selector omits `[contenteditable]`, `summary`, `audio[controls]`, `video[controls]`, and anything focused inside a `tabindex="-1"` region. Shift+Tab from one of those then jumps to the **last** control in the dialog instead of moving back one — worse than not trapping, because the wrap is silent.

Not reachable through today's modal bodies (inputs, selects and buttons only), which is why no test catches it. It becomes reachable the moment a body grows a rich-text field or a `<details>`.

Separating the two costs one line: treat `-1` as "the container" only when `document.activeElement === dialog`, and otherwise leave the event alone.

backend/src/main/java/com/healthupgrades/healtharea/adapter/in/web/HealthAreaRequest.java
```diff
@@ -1,24 +1,34 @@
...
         String description,
         Integer priority,
-        String icon,
-        String color
-) {}
+        @Size(max = ICON_MAX) String icon,
+        @Size(max = COLOR_MAX) String color
+) {
+    /** Mirrors {@code health_areas.name VARCHAR(255)}; see BR-16 for why the bound is here. */
+    static final int NAME_MAX = 255;
```
NaimElijah: **Nothing keeps these constants and the columns they mirror in step.**

`NAME_MAX`/`ICON_MAX`/`COLOR_MAX` here, plus `TITLE_MAX`, `UNIT_MAX`, `TARGET_UNIT_MAX` and `AuthController`'s `NAME_MAX`/`EMAIL_MAX`, each restate a number that lives in `V1__init_schema.sql`. A later migration that narrows `health_areas.icon` leaves `@Size(max = 100)` behind, and the result is exactly the `DataIntegrityViolationException` → 500 that BR-16 exists to prevent — reintroduced silently, with the requirement still claiming it cannot happen. The entities declare no `@Column(length = …)` either, and Hibernate's `validate` does not check column length, so nothing in the build notices.

The PR already names the right pattern in its own trade-offs: `FrontendEnumContractTest`. One test that reads the column lengths and compares them against the DTO bounds would make BR-16 checkable rather than trusted — which is the standard `requirements.md` sets for itself.

backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
```diff
@@ -76,9 +99,9 @@ private UserDto toUserDto(User user) {
 
     /** Registration request body. */
     public record RegisterRequest(
-            @NotBlank String name,
-            @NotBlank @Email String email,
-            @NotBlank String password
+            @NotBlank @Size(max = NAME_MAX) String name,
+            @NotBlank @Email @Size(max = EMAIL_MAX) String email,
+            @NotBlank @Size(min = PASSWORD_MIN, max = PASSWORD_MAX) String password
```
NaimElijah, line 104: **With `@NotBlank` and `@Size(min = 8)` on the same field, the 400 message for an empty password is now non-deterministic.**

`GlobalExceptionHandler.handleMethodArgumentNotValid` builds the field map with `errors.put(fe.getField(), fe.getDefaultMessage())` (line 177), so a field carrying two violations keeps whichever one `getFieldErrors()` happens to return last — and Hibernate Validator does not specify that order. `POST /api/auth/register` with `"password": ""` violates both constraints, so the caller may be told `"must not be blank"` on one request and `"size must be between 8 and 72"` on an identical one.

`GivenABlankPassword_…` only asserts that `fieldErrors.password` exists, so it cannot see the flip.

Collecting the messages per field (a list, or joined) fixes it; leaving it means a documented wire contract that varies run to run.

docs/requirements/requirements.md
```diff
@@ -109,7 +109,7 @@ nothing is shared between accounts.
 | FR-37 | The theme control is reachable before signing in | `LoginPage.test.tsx` |
 | FR-38 | A page's content grows with the browser window, up to one cap chosen for readability | `PageContainer`, `PageContainer.test.tsx` |
 | FR-39 | A dialog fits the window at any size: its heading stays put and only its body scrolls | `Modal` — checked by hand, see §6 |
-| FR-40 | A dialog announces itself as a dialog named by its title | `Modal`, `Modal.test.tsx` |
+| FR-40 | A dialog announces itself as a dialog named by its title, takes focus on open, confines Tab to its own controls, restores focus on close, and marks the page behind it inert | `Modal`, `Modal.test.tsx`, [ADR-013](../ADRs/ADR-013-trapping-focus-without-a-native-dialog.md) |
```
NaimElijah, line 112: **`docs/architecture/architecture.md` is not updated, and ADR-013 is not in its decision index.**

The user-level `CLAUDE.md` requires `architecture.md` to carry "**Structural decisions not visible from the file layout** … anything a maintainer would otherwise have to reverse-engineer", updated "in the same commit as the change". This PR introduces the SPA's first React portal and a module-level refcounted mutation of `document.body`'s children — a global invariant constraining every future component that wants to portal — and `architecture.md` says nothing about either.

Its "Where the reasoning lives" table also still carries only

> Why every page shares one width; why the shell can be trusted not to overflow; why container queries and a native `<dialog>` were turned down — ADR-005

ADR-013 re-decides that `<dialog>` question on new grounds and supersedes ADR-005 in part, so a reader the table sends to ADR-005 lands on the half that is no longer true.

backend/src/main/java/com/healthupgrades/auth/adapter/in/web/AuthController.java
```diff
@@ -23,6 +24,28 @@
 @RequiredArgsConstructor
 public class AuthController {
 
+    /** Mirrors {@code users.name VARCHAR(255)}; see BR-16 for why the bound is here. */
+    static final int NAME_MAX = 255;
```
NaimElijah: **Package-private with no consumer outside the file.**

`NAME_MAX`, `EMAIL_MAX`, `PASSWORD_MIN`, `PASSWORD_MAX` (and `TITLE_MAX`, `UNIT_MAX`, `TARGET_UNIT_MAX`, `HealthAreaRequest`'s three) are `static final` rather than `private static final`, which normally signals "a test reads this". None does: `AuthControllerTest` writes `"p".repeat(73)` and `"12345678"`, and the controller tests write `256`, `101`, `51`, `255` as literals.

So the widened visibility buys nothing, and the duplicated literals are what a future bound change has to be chased through. Either make them `private`, or have the boundary tests reference them so the cases move with the constant.

frontend/src/components/ui/Modal.tsx
```diff
@@ -49,15 +104,75 @@ const Modal: React.FC<ModalProps> = ({ isOpen, onClose, title, children }) => {
...
+
+    const previouslyFocused = document.activeElement as HTMLElement | null;
+    acquireInert();
+    dialogRef.current?.focus();
+
+    return () => {
+      // Order matters: a browser refuses to move focus into an inert subtree, so the page has to be
+      // released before the trigger can take focus back. Keeping both in one effect makes that
+      // explicit — split in two it would rest on declaration order, and jsdom would never show the bug.
+      releaseInert();
```
NaimElijah, line 151: **With two dialogs closing in the same commit, this ordering guarantee does not hold.**

The comment is right that the page has to be released before the trigger can take focus back. But `releaseInert()` short-circuits while `openModals > 0`, so when an outer dialog closes with an inner one still open — or when both unmount together, as on a route change — `previouslyFocused.focus()` runs against an element still inside the `inert` subtree. A browser silently ignores that, focus falls to `<body>`, and the escape hatch this effect exists to provide is gone. jsdom implements neither half, so nothing here can show it.

Guarding the restore on `!previouslyFocused.closest('[inert]')`, or deferring it until the count actually reaches zero, would make the stated invariant true rather than true-in-the-common-case.

frontend/src/components/ui/Modal.test.tsx
```diff
@@ -40,12 +107,38 @@ describe('Modal', () => {
...
+
+    // Pins the attribute, not the behaviour — jsdom enforces no inertness whatsoever. RTL's own
+    // container is a direct child of <body>, which is exactly what the walk marks.
+    it('GivenAnOpenModal_WhenItRenders_ThenTheContentBehindIsMarkedInert', () => {
+      const { container } = renderModal();
+
+      expect(container.getAttribute('inert')).toBe('');
+    });
+
+    it('GivenAModalThatHasClosed_WhenItUnmounts_ThenTheContentBehindIsNoLongerInert', () => {
```
NaimElijah: **The name says `WhenItUnmounts`, but the action is Escape and the modal does not unmount.**

`Openable` keeps the dialog mounted and flips `isOpen` to false; `Modal` then returns `null` before the portal, which is a close, not an unmount. The Given/When split is inverted against ADR-009's `Given<state>_When<action>_Then<outcome>` too: "a modal that has closed" is the outcome, and "unmounts" is not the action performed.

`GivenAnOpenModal_WhenItCloses_ThenTheContentBehindIsNoLongerInert` says what the test does — and leaves the name free for a real unmount test, which is a genuinely different path through the effect cleanup.

frontend/src/components/ui/Modal.test.tsx
```diff
@@ -40,12 +107,38 @@ describe('Modal', () => {
...
+
+      fireEvent.keyDown(document, { key: 'Escape' });
+
+      expect(container.getAttribute('inert')).toBeNull();
+    });
+
+    // The overlays skip each other by attribute rather than by a registry built in an effect, because
+    // React inserts both portals before it runs either effect.
+    it('GivenTwoOpenModals_WhenTheyRender_ThenNeitherMarksTheOtherInert', () => {
+      render(<TwoModals />);
```
NaimElijah, line 145: **`TwoModals`' comment claims a case `HealthAreasPage` does not produce.**

> Two modals mounted at once, which is what `HealthAreasPage` does with its create/edit/delete trio.

`HealthAreasPage` mounts three `Modal` *components*, but `isCreateOpen`, `editArea` and `deleteId` are never set together — the cards that would set a second one are behind the backdrop and, since this PR, `inert`. A closed `Modal` returns `null` before the portal and its effect returns early, so it is not "mounted" in the sense the refcount cares about.

The refcount is defensive machinery for a state the app cannot currently reach, which is a defensible call — but the comment presents it as covering a live case and the next reader will trust it.

NaimElijah:
## Review response — 10 of 12 built, 2 recorded

All twelve were real. Thanks in particular for the first one; it was the finding that mattered most,
because it meant the PR's own justification was false.

### Built

**1 — the 400s never reached a form.** Correct, and it undercut the premise of the first commit: the
API started naming the field and every modal form threw the answer away, with `await
mutateAsync(...)` producing an unhandled rejection and a dialog that sat there. Added `toFormMessage`
beside `toApiError` — it prefers the field detail over the generic "Validation failed" a 400 puts in
`message` — and wired the backlog create, health-areas create and edit, and register paths to it.
Six tests on the decoder.

**2 — RegisterPage rendered `message`.** Same fix. A pasted passphrase past BCrypt's 72 bytes now
says what is wrong instead of "Validation failed (reference …)".

**5 — `index <= 0` conflated two states.** Right, and the doc comment claimed the -1 branch was
deliberate when it only handled one of the two things that produce it. Told apart by identity now,
and "focus is on something the list does not know about" makes the handler stand aside rather than
guess a boundary. `FOCUSABLE` also grew the omissions you named. New test, and it fails against the
old branch.

**6 — nine constants restated the migrations with nothing enforcing agreement.** The strongest
structural finding. `ColumnBoundContractTest` now reads the migrations and fails on a divergence, in
the shape of `FrontendEnumContractTest`; it also fails when a *new* bounded column is neither
mirrored by a DTO nor listed as one no request body writes to, so the next one has to be considered.
Verified by mutation — narrowing `ICON_MAX` to 99 fails it with the reason.

**7 — non-deterministic message for a doubly-violated field.** Confirmed at
`GlobalExceptionHandler:175`. Sorted, first-wins, with a test that repeats the call because one
attempt cannot tell a stable choice from a lucky one.

**8 — architecture.md and the ADR index.** A straight miss against the repo's own rule. It now carries
the portal, the body-level `inert` invariant, the two constraints that puts on anything added later,
and the hand-copied-bounds contract. ADR-013 is in the index; ADR-005's row no longer claims the
`<dialog>` half.

**9 — visibility with no consumer.** Fixed in the opposite direction: the contract test lives in
another package, so the constants are now `public` with a real reader, and the controller tests use
them instead of repeating 256/101/51/73.

**10, 4, 3 — three comments claiming more than the code does.** All correct. The release-before-restore
ordering holds only on the last close; the inert set is snapshotted once, so a later portal is never
marked; Escape is on `document` and closes every open dialog, not the topmost. Each is now stated
where it applies, and ADR-013 carries the first two.

**11, 12 — two tests describing things that do not happen.** The unmount test tested a close, renamed.
The `TwoModals` comment claimed `HealthAreasPage` produces two open dialogs; it cannot, and the
comment now says so — the refcount is defence for an unreachable state, which is the honest reason for
it in a shared primitive that does not control its call sites, rather than cover for a live path.

### Recorded, not built

**1, in part — `UpgradeDetailsPage`'s four `mutate(...)` forms.** They have no error state at all, so
this is adding one to the component and a slot to each of four dialogs, not stopping a discard. Filed
as #76, which also raises whether the slot belongs in `Modal` now that six forms do this by hand.

**The toast interaction.** Filed as #74 before the review, and it is the same call: correct under
`aria-modal`, confusing in practice, and a second component's problem.

Also filed while verifying: #73 (Spring Boot 3.2 is eight months past extended EOL, which is what
blocks #59) and #75 (the frontend container's healthcheck probes `localhost`, which resolves to `::1`
while nginx binds IPv4 — it can never pass).

### Verification after the fixes

`mvn verify` 485 + 39 green; `npm run test` 147 green; lint, `check:colours`, `build` and the diagram
drift check all clean.

